PR #6, opened 2022-02-23
feat(package): rename packages

nichonien:
This PR updates the repo and package names to be more generic as mentioned below.

iam-roles               -> ew-credentials
onchain-role-enrolment  -> ew-onchain-claims
role-governance         -> ew-credential-governance
role-vc-verification    -> ew-vc-verification


package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@energyweb/iam-roles",
+  "name": "ew-credentials",
```
jrhender, line 2: IMO, it still makes sense to have the packages namespaced under `@energyweb`. What do you think @nichonien ?
```suggestion
  "name": "@energyweb/ew-credentials",
```

jrhender (changes requested):

package.json
nichonien: @jrhender I thought of it but then `@energyweb/ew-credential-governance` is a long package name. Also, `ew` does resemble `energyweb` in a shorter way/form.

jrhender: Hmm, IMO, all of energyweb's packages should be scoped under `@energyweb`. I can see how it seems a bit redundant but one could argue that `@energyweb` refers to the organization publishing the package and `ew` refers to the fact that we are doing credential governance in our more specific way.
If I had to chose between `@energyweb` and `ew`, I would chose `@energyweb`. So maybe the package name should be `@energyweb/credentials`. @mhrsntrk @jcmelati what are your thoughts here?

jrhender: Ok, after discussing with @mhrsntrk in standup, we will proceed with `@energyweb/credentials`

PR #8, opened 2022-03-02
feat(workflow): add deploy workflow for publishing

nichonien:
This PR adds deploy workflow for publishing the packages.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -34,15 +34,6 @@ jobs:
           node-version: '16.10.0'
           registry-url: https://registry.npmjs.org/
 
-      - name: Configure Git
-        run: |
-          git config --global user.email ${{ secrets.GH_EMAIL }}
-          git config --global user.name ${{ secrets.GH_USER }}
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,102 @@
...
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+          ref: develop
+          token: ${{ secrets.GH_TOKEN }}
+
+      - name: Commit to develop branch and merge to master branch
+        run: |
+          git config --global user.email ${{ secrets.GH_EMAIL }}
```
JGiter, line 98: I am not sure if we need to set email and name since `develop` has already been checkout with `GH_TOKEN`

JGiter (approved):

.github/workflows/deploy.yml
nichonien: A token is associated with a user and they might have different emails / orgs, so this is needed IMO.

JGiter (approved):

PR #10, opened 2022-03-04
feat(claims): remove deprecated `fields` property

Passerino:
BREAKING CHANGE: deprecated `fields` property removed. Use `requestorFields` instead.


nichonien (approved):
Looks good, thanks @Passerino !

PR #13, opened 2022-03-25
Verify issuers chain of trust

nichonien:
This PR contains code to verify chain of trust for holder's credential.


packages/vc-verification/src/issuer-verification.ts
```diff
@@ -0,0 +1,146 @@
...
+            dids[i],
+            this.ASSERTION_DELEGATE_TYPE,
+            issuerDID
+          ))
+        ) {
+          didMatched = true;
+          break;
+        }
+      }
+    }
```
jrhender, line 122: I don't think this query of the DID registry should be handled here. It should be handed by the did:ethr resolver (e.g. https://github.com/decentralized-identity/ethr-did-resolver/blob/2a66a2bfc29ebb59a33967f6dec8cdc676c2fee5/src/resolver.ts#L164) and the delegate will just be included in the DID document

jrhender (changes requested):
Thanks for the initial PR @nichonien . I added some suggestions

packages/vc-verification/src/issuer-verification.ts
```diff
@@ -0,0 +1,146 @@
...
+  private ASSERTION_DELEGATE_TYPE =
+    '0x766572694b657900000000000000000000000000000000000000000000000000';
+
+  /**
+   *
+   * @param signer
+   * @param claimManagerAddr
+   */
+  constructor(
+    signer: JsonRpcSigner,
```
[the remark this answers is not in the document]
nichonien: We need a `signer` to connect with `RoleDefResolver` and also the `signer` would guarantee `provider` for initialising `Resolver` I guess.

packages/vc-verification/src/issuer-verification.ts
```diff
@@ -0,0 +1,146 @@
...
+    roleDefResolverAddr: string,
+    registrySetting: RegistrySettings,
+    ipfsUrl: string
+  ) {
+    this._signer = signer;
+    this._ipfsUrl = ipfsUrl;
+    this._ipfsStore = new DidStore(this._ipfsUrl);
```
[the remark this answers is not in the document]
nichonien: >The verifyChainOfTrust function can then rely on the accreditations of the resolved issuer
The accreditation has to come from a governance framework which we don't have as of now. The `issuerResolver` functionality would need to fetch the trusted authoritative credential from IPFS for our case, we could separate it from verification code though. Also, we agreed that the proposed governance [template](https://energyweb.atlassian.net/wiki/spaces/MYEN/pages/3071246337/VC+Verification#Sample-Governance-Template) is a nice way to go but this would have added more complexity and efforts which we don't want for now. Thus routing for issuer DID resolution and accessing credential from IPFS.

packages/vc-verification/src/issuer-verification.ts
```diff
@@ -0,0 +1,146 @@
...
+    );
+    let didMatched = false;
+    if (dids.length > 0) {
+      for (let i = 0; i < dids.length; i++) {
+        if (
+          dids[i] == issuerDID ||
+          (await this._didRegistry.validDelegate(
+            dids[i],
+            this.ASSERTION_DELEGATE_TYPE,
+            issuerDID
```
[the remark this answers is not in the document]
nichonien: Yeah, you are right and `issuerDID` also needs to be address.

packages/vc-verification/src/issuer-verification.ts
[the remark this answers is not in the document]
jrhender: Are you sure we need a `signer` to connect to `RoleDefResolver`?

Yes, one can get a `provider` from a `signer`, but we should let that happen at instantiation time. In others words, we should just accept a `Provider` in the constructor.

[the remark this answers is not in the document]
nichonien: Hmm, you are right It should be through providers.

packages/vc-verification/src/issuer-verification.ts
jrhender: Ah, I think maybe there is confusion on how delegates for `did:ethr`. I could be wrong, but as I see it delegates in `did:ethr` are addresses that are allowed to sign on behalf of the DID, they are not addresses that are equivalent identifiers to the DID.

In other words, if the issuer of a credential is a delegate of a valid issuer according to the role definition, this does not make the issuer of the credential valid.

For example, if the role definition specifies that `did:123` the valid issuer, then this is a valid credential:
```js
{
    issuer: "did:123",
    credentialSubject: { ... },
    proof: {
      verificationMethod: "did:123#delegate-did:456"
      ....
   }
```
whereas this is not:
```js
{
    issuer: "did:456",
    credentialSubject: { ... },
    proof: {
      verificationMethod: "did:456#controller"
      ....
   }
```

In other words, if did:456 wants to issue a credential on behalf of did:123, we should expect that did:456 should still put did:123 as the `issuer` in the credentials that they are issuing.
This is more correct in my view and has the benefit of allowing the code to remain generic to the general DID spec and not coupled to did:ethr in anyway.

packages/vc-verification/src/issuer-verification.ts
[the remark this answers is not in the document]
jrhender: Suggested change is here: https://github.com/energywebfoundation/ew-credentials/pull/14

packages/vc-verification/src/issuer-verification.ts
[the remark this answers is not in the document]
jrhender: Yes, I think that separating it from the verification good is a step in the right direction 👍 

> The issuerResolver functionality would need to fetch the trusted authoritative credential from IPFS for our case, we could separate it from verification code though

packages/vc-verification/src/issuer-verification.ts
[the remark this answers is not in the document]
nichonien: @jrhender I had mode those changes in this PR, so closing #14 for now.

packages/vc-verification/src/issuer-verification.ts
```diff
@@ -0,0 +1,148 @@
...
+      const service = subjectDIDDoc.service;
+      let serviceEndpoint = '';
+      for (const sv of service) {
+        if (utils.namehash(sv.type) === role) {
+          serviceEndpoint = sv.serviceEndpoint;
+          hasParent = false;
+          break;
+        }
+      }
+      const issuerDID = await this.verifyCredential(serviceEndpoint);
```
jrhender, line 73: Here's some suggested pseudo code on how I think this could be make more flexible and generic. It probably isn't perfect but let me know what you think.

For instance, we shouldn't have this logic hardcoded to specific verification code. It would be better to pass a `credentialVerifyCallback` so we can update the verification (e.g. to use DIDKit credential verification) without needing to change this code.
```suggestion
  async verifyChainOfTrust(credential: IVerifiableCredential, verifyCredentialProofCallback: (credential: IVerifiableCredential) => Promise<VerificationResult>) {
    let hasParent = true;
    while (hasParent) {
      const role = parseCredentialForRole(credential);
      const roleDefinition = resolveDefinition(role);
      if (roleDefinition.didIssuer) {
        // check if credential.issuer is in roleDefinition
      }
      else {
         const issuerCredentials = getIssuer(issuerDid).accreditations // ideally, getIssuer should be VUI issuer interface
         const correctCredential = findCredential(issuerCredentials, roleDefintion.requiredRole)
         const proofVerificationResult = await verifyCredentialProofCallback(credential);
         // Repeat the process to get authorization of the issuer of the credential...
      }
```

nichonien: Hmm, sounds good. Let me put more thoughts into it.

nichonien: @jrhender I think we can have two separate APIs for this verification. First being the hardcoded way which is the current implementation and the second with the `credentialVerifyCallback` which consumes any external library.

The first/current verification API relies on the the `role` and holder's `did`, this way it interacts with the source of truths( credential storage, DIDDoc) and verifies the chain of trust. 

While the `credentialVerifyCallback` would be a general VC verification approach, which may and mayn't rely on our `role` based credential ecosystem and can verify the same.

jrhender: Ok, we can give it a try to see how it looks 👍   
I'm just looking ahead and I think we're very soon going to want to use DIDKit verification instead of (or at least in addition to) `ProofVerifier` but I'm open to whatever way you think is best to do that.

packages/vc-verification/src/issuer-verification.ts
```diff
@@ -0,0 +1,255 @@
...
+  constructor({
+    provider,
+    roleDefResolverAddr,
+    registrySetting,
+    storageSettings,
+  }: {
+    provider: providers.Provider;
+    roleDefResolverAddr: string;
+    registrySetting: RegistrySettings;
+    storageSettings: StorageSettings;
```
jrhender, line 36: Maybe consider using name of `credentialResolverSettings` to match the `CredentialResolver` class where this param is used.
```suggestion
    storageSettings: StorageSettings;
```

packages/vc-verification/src/issuer-verification.ts
```diff
@@ -0,0 +1,255 @@
...
+  }) {
+    this._provider = provider;
+    this._registrySetting = registrySetting;
+    this._roleDefResolver = RoleDefinitionResolverV2__factory.connect(
+      roleDefResolverAddr,
+      this._provider
+    );
+    this._resolver = new Resolver(this._provider, this._registrySetting);
+    this._credentialResolver = new CredentialResolver(storageSettings);
```
jrhender, line 45: Instead of hardcoding the `CredentialResolver` construction, I think it would be better to accept the `CredentialResolver` object directly. Or even better would be to accept an `ICredentialResolver` interface so that we could swap out the implementation in the future. But if that's overkill for now, no worries.

jrhender (changes requested):
I think it's really coming together, looks great! Some suggestions for improvement though.

packages/vc-verification/src/issuer-verification.ts
```diff
@@ -0,0 +1,255 @@
...
+   * @param did fetches serviceEndpoint from the subject's DIDDoc
+   * @param role
+   * @returns
+   */
+  async getCredentialUrl(did: string, role: string) {
+    const subjectDIDDoc = await this._resolver.read(did);
+    const service = subjectDIDDoc.service;
+    let serviceEndpoint = '';
+    for (const sv of service) {
+      if (utils.namehash(sv.type) === role) {
```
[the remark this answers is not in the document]
nichonien: I guess so, the services in the DIDDoc has `type` as one of the attributes that stores the namespace / role. This is the [interface](https://github.com/energywebfoundation/ew-did-registry/blob/0ab6e12ee011b4c2e6fdd0d2927d7f757fd54d64/packages/did-resolver-interface/src/models/resolver.ts#L38) declaration.

packages/vc-verification/src/issuer-verification.ts
```diff
@@ -0,0 +1,255 @@
...
+      return 'Insufficient information';
+    }
+  }
+
+  /**
+   * Verifies chain of trust for a given holder's DID and role
+   * @param {string} subjectDID
+   * @param {string} role
+   */
+  async verifySubjectChainOfTrust(subjectDID: string, role: string) {
```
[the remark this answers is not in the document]
nichonien: ok, do we need to support `LegacyEWCredential` within this verification code. I see this `verification` functionality been consumed in our VC presentation and other verifiable credential operations as per latest specs, I think it's better to enforce the latest changes and restrict it to verifiable credential only. Since the legacy credential already has verification mechanism and we would eventually discard them. What do you think?

packages/vc-verification/src/issuer-verification.ts
```diff
@@ -0,0 +1,255 @@
...
+      }
+    }
+  }
+
+  /**
+   * Fetches role form a credential
+   * @param credential
+   * @returns
+   */
+  async parseRoleFromCredential(credential: IVerifiableCredential) {
```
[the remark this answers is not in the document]
nichonien: Hmm, IMO only `parseRoleFromCredential` can be a private method, rest are more of utility functions.

packages/vc-verification/src/issuer-verification.ts
nichonien: I guess the `CredentialResolver` object seems to be reasonable as one could have flexibility to decide the approach / configuration for resolution of credential, it could be either from IPFS or some hosted / published accreditation based on governance. So it would be up to the caller or the need of the resolution to decide on the way of initialisation of `CredentialResolver` and changes could be made directly to `CredentialResolver` as per need, while `verification` code remains intact.

packages/vc-verification/src/issuer-verification.ts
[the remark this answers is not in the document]
jrhender: Yea, I definitely agree that we should have been setting a type on the service. It's required by the DID Spec. https://www.w3.org/TR/did-core/#services
But I'm 99% percent sure that we aren't doing it, it is a bug in our system.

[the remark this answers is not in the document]
nichonien: Was there any reason for not setting the `type` attribute with namespace or we missed to do so? Also, I see in all the tests be it `ew-did-registry` or `sis-hub` the attribute has been set in those tests so I assumed it is adopted from the specs. 

Also, if the `type` attribute is not being set, how do we know the namespace from the service, except from the claimData? Like if I resolve my DIDDoc, I could not get the namespace then without decoding the token / serviceEndpoint or am I missing something?

packages/vc-verification/src/models/credential-resolver.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * Storage settings to fetch issuer's credential
+ */
+export interface CredentialResolverSettings {
+  ipfsApi: string;
```
jrhender, line 5: Minor point but I think `Url` is a bit more accurate and in line with what we use elsewhere
```suggestion
  ipfsUrl: string;
```

packages/vc-verification/src/models/verifiable-credentials.ts
```diff
@@ -0,0 +1,37 @@
+export interface IVerifiableCredential {
```
jrhender, line 1: Can this be imported from Jakub's new ew-did-registry package instead of duplicated?

packages/vc-verification/src/issuer-verification.ts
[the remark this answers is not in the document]
jrhender: I think we just missed to do so, unless you're aware of a reason @JGiter .

>  if the type attribute is not being set, how do we know the namespace from the service, except from the claimData? Like if I resolve my DIDDoc, I could not get the namespace then without decoding the token / serviceEndpoint or am I missing something?

Yes, I think that is the case (unfortunately), that one needs to retrieve and decode the token in order to get the namespace

packages/vc-verification/src/issuer-verification.ts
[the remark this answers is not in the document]
nichonien: Yeah, it would be a good approach and then it could be deprecated whenever needed.

packages/vc-verification/src/models/credential-resolver.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * Storage settings to fetch issuer's credential
+ */
+export interface CredentialResolverSettings {
```
jrhender, line 4: This file should be renamed to `credential-resolver-settings.ts`, I believe

packages/vc-verification/src/issuer-credential-resolver.ts
```diff
@@ -0,0 +1,23 @@
+import { StorageSettings } from './models';
+import { DidStore } from '@ew-did-registry/did-ipfs-store';
+import { IDidStore } from '@ew-did-registry/did-store-interface';
+
+export class CredentialResolver {
```
jrhender, line 5: I think file can be renamed to just `credential-resolver` to match class name

packages/vc-verification/src/issuer-credential-resolver.ts
```diff
@@ -0,0 +1,23 @@
...
+    this._ipfsUrl = storageSettings.ipfsApi;
+    this._ipfsStore = new DidStore(this._ipfsUrl);
+  }
+
+  /**
+   *
+   * @param serviceEndpoint
+   * @returns {string}
+   */
+  async getIssuerCredential(serviceEndpoint: string): Promise<string> {
```
jrhender, line 19: I think it would be good to make this resolution more robust by copying the code from passport-did-auth. It has some code to deal with edge cases in legacy credentials https://github.com/energywebfoundation/passport-did-auth/blob/develop/lib/LoginStrategy.ts#L267
What do you think?

nichonien:
@jrhender I have been thinking of the using `cache` d `RoleDefinition` for the verification. As we discussed that some cache interface could be put to use which receives the cached definitions as part of verification code, but I'm not sure how we get the same for the roles somewhere in the upper hierarchy.

So for `holder -> issuer1 -> issuer2 ->issuer3`, we can very well fetch the roleDefinition for holder's credential and send it, but for any other issuers like issuer2, the caller (iam-lib / ssi-hub) needs to resolve the complete chain and then send the data. This seems to be some good processing and I could think of it as almost equivalently efficient as that of fetching form the blockchain. What are your thoughts?

packages/vc-verification/src/issuer-verification.ts
```diff
@@ -0,0 +1,234 @@
...
+   */
+  constructor({
+    provider,
+    roleDefResolverAddr,
+    registrySetting,
+    credentialResolver,
+  }: {
+    provider: providers.Provider;
+    roleDefResolverAddr: string;
+    registrySetting: RegistrySettings;
```
jrhender, line 38: Similar to how we accepting a `CredentialResolver`, I think that `IssuerVerification` should accept a `RoleDefinitionResolver` so that it is independent of role definition resolution implementation.

packages/vc-verification/src/credential-resolver.ts
```diff
@@ -0,0 +1,111 @@
...
+import { CredentialResolverSettings } from './models';
+import { DidStore } from '@ew-did-registry/did-ipfs-store';
+import { IDidStore } from '@ew-did-registry/did-store-interface';
+import { OffChainClaim } from './models';
+import { isValidErc1056, getDIDChain } from '@ew-did-registry/did';
+import { Resolver } from '@ew-did-registry/did-ethr-resolver';
+import { RegistrySettings } from '@ew-did-registry/did-resolver-interface';
+import * as jwt from 'jsonwebtoken';
+
+export class CredentialResolver {
```
jrhender, line 11: I think it makes sense to be a bit more specific with the class name as I predict that we will have other types of `CredentialResolvers` in the future.
E.g. I anticipate having a `VuiIssuerApiCredentialResolver` in the future (https://gataca-io.github.io/vui/issuerResolution#resolution-api)
My suggested name might be a bit wordy, but let me know what you think.
```suggestion
export class DidServiceIpfsCredentialResolver {
```

jrhender (commented):
@nichonien To answer your question, IMO, we should use the same pattern for role definition resolution that we are using for credential resolution. I've come to realize that what I am recommending is the "Strategy" pattern.

In other words, I think that the `IssuerVerification` class should accept an abstract `CredentialResolutionStrategy` and an abstract `RoleDefinitionResolutionStrategy`. We can then have different implementations of the strategy. 
So in the case of cached data from the ssi-hub cache, a `CachedRoleDefinitionResolutionStrategy` can be implemented and passed to the `IssuerVerification` class.

packages/vc-verification/src/credential-resolver.ts
nichonien: The `CredentialResolver` could be the upfront class which manages / instantiates the specific resolvers like `DidServiceIpfsCredentialResolver` and `VuiIssuerApiCredentialResolver`. Thus the underlying approach would be abstracted by `CredentialResolver` and we can have as many resolver approaches as we want. The `CredentialResolverSettings` could handle the settings for different type of resolution options and be provided to `CredentialResolver`. What do you think?

packages/vc-verification/src/issuer-verification.ts
nichonien: Hmm, sounds good. Or maybe we can do `CachedObjectStrategy` which can accept both `CachedRoleDefinitionResolutionStrategy` and `CachedDidDocumentResolutionStrategy` and resolves the passed objects. Does this sound right?

packages/vc-verification/src/credential-resolver.ts
jrhender: I think it is best if the client can pass their own implementation, without needing to modify `CredentialResolver`. It is not good if `CredentialResolver` needs to be modified for every new "type" of credential resolver.
I'm open to having `CredentialResolver` being a factory which instantiates the various types of resolvers. If you like that idea, I would change the name to `CredentialResolverFactory` and have it accept the settings object you describe and have it return `DidServiceIpfsCredentialResolver`. Personally, I don't think it is really necessary to have such a factory at this point. I personally don't like the idea of a single `CredentialResolverSettings` type that is used for both `DidServiceIpfsCredentialResolver` and `VuiIssuerApiCredentialResolver` as it couples the construction of the various types of `CredentialResolver`.

packages/vc-verification/src/issuer-verification.ts
jrhender: Personally I don't think having `CachedObjectStrategy` is a good idea because it needlessly couples  `CachedRoleDefinitionResolutionStrategy` and `CachedDidDocumentResolutionStrategy`. To me, `CachedObjectStrategy` implies that it is an implementation of `AbstractObjectStrategy` (to use the strategy pattern) and I think `AbstractObjectStrategy` is overly abstract. In other words, an "Object Strategy" is too broad in my view.

nichonien: Hmm, makes sense. I think we now have a better understanding of what we want.

packages/vc-verification/src/credential-resolver.ts
nichonien: Ok, separate implementation of resolver being passed by the client as per their choice and `CredentialResolver` able to verify them sounds right.

packages/vc-verification/src/role-definition-resolver.ts
```diff
@@ -0,0 +1,26 @@
...
+  /**
+   * The RoleDefinitionResolver class could be initialised either with
+   * CachedRoleDefinitionResolver or ENSRoleDefinitionResolver
+   * @param rolDefResolver
+   */
+  constructor(
+    rolDefResolver: CachedRoleDefinitionResolver | ENSRoleDefinitionResolver
+  ) {
+    this._roleDefResolver = rolDefResolver;
+  }
```
jrhender, line 16: Would clients need to modify this code in order to add a new implementation?
I'm wondering if maybe it would be better if `RoleDefinitionResolver` were an interface so that a client (e.g. ssi-hub`) can add their own implementation without needing to modify this code.

packages/vc-verification/src/role-resolvers.ts
```diff
@@ -0,0 +1,66 @@
...
+  async getRoleIssuers(namespace: string): Promise<IIssuerDefinition> {
+    let issuers: IIssuerDefinition = {};
+    for (let role of this._roleDefinitions) {
+      if (role.roleName == namespace) {
+        issuers = role.issuer;
+      }
+    }
+    return issuers;
+  }
+}
```
jrhender, line 35: I don't think it's necessary for ew-credentials to provide an implementation of `CachedRoleDefinitionResolver` as PDA/ICL/ssi-hub, will provide their own implementations as necessary.

packages/vc-verification/src/role-resolvers.ts
```diff
@@ -0,0 +1,66 @@
...
+  IIssuerDefinition,
+  IRoleDefinitionV2,
+} from '@energyweb/credential-governance';
+import { providers } from 'ethers';
+import { RoleDefinitionResolverV2__factory } from '@energyweb/credential-governance/ethers/factories/RoleDefinitionResolverV2__factory';
+import { RoleDefinitionResolverV2 } from '@energyweb/credential-governance/ethers/RoleDefinitionResolverV2';
+
+abstract class RoleResolver {
+  abstract getRoleIssuers(namespace: string): Promise<IIssuerDefinition>;
+}
```
jrhender, line 11: In general, I think it is better to use interfaces rather than abstract classes as a class can implement multiple interfaces.

Maybe call it `IssuerDefinitionResolver` as it looks like it is IssuerDefinitions that are being resolved?

```suggestion
interface class IssuerDefinitionResolver {
  getRoleIssuerDefinition(namespace: string): Promise<IIssuerDefinition>;
}
```

packages/vc-verification/src/Ipfs-credential-resolver.ts
```diff
@@ -0,0 +1,81 @@
...
+    const offChainClaims = await this.offchainClaimsOf(did);
+    for (const claim of offChainClaims) {
+      if (claim.claimType === role) {
+        return claim;
+      }
+    }
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  async isOffchainClaim(claim: any): Promise<boolean> {
```
jrhender, line 42: Maybe this can be `private`

packages/vc-verification/src/Ipfs-credential-resolver.ts
```diff
@@ -0,0 +1,81 @@
...
+      'claimType',
+      'claimTypeVersion',
+      'issuedToken',
+      'iss',
+    ];
+    const claimProps = Object.keys(claim);
+    return offChainClaimProps.every((p) => claimProps.includes(p));
+  }
+
+  async offchainClaimsOf(did: string): Promise<OffChainClaim[]> {
```
jrhender, line 53: Maybe this can be `private`

packages/vc-verification/src/credential-resolver.ts
```diff
@@ -0,0 +1,17 @@
+import { OffChainClaim } from './models';
+
+/**
+ * An interface for a credential resolver
+ */
+export interface CredentialResolver {
+  /**
+   *
+   * @param did
+   * @param namespace
```
jrhender, line 10: Would be good to add a bit more documentation about the parameters here

packages/vc-verification/src/issuer-def-resolver.ts
```diff
@@ -0,0 +1,44 @@
...
+import { providers } from 'ethers';
+import { RoleDefinitionResolverV2__factory } from '@energyweb/credential-governance/ethers/factories/RoleDefinitionResolverV2__factory';
+import { RoleDefinitionResolverV2 } from '@energyweb/credential-governance/ethers/RoleDefinitionResolverV2';
+
+/**
+ * An interface for Resolution of issuers for a namespace
+ */
+export interface IssuerDefinitionResolver {
+  /**
+   * @param namespace
```
jrhender, line 11: Would be good to add a bit more documentation

packages/vc-verification/src/issuer-def-resolver.ts
```diff
@@ -0,0 +1,44 @@
...
+   * @returns IIssuerDefinition for the given namespace
+   */
+  getRoleIssuerDefinition(namespace: string): Promise<IIssuerDefinition>;
+}
+
+export class RoleIssuerDefinitionResolver implements IssuerDefinitionResolver {
```
jrhender, line 17: IMO, this name could be improved. I would add something about how it is resolving using an ethers provider (to contrast it with a resolver from cached data, for example). Name is bit clunky but I think 
```suggestion
/**
 * Resolves an issuer definition by reading smart contract state via an Ethers provider
 */
export class EthersProviderIssuerDefinitionResolver implements IssuerDefinitionResolver {
```

packages/vc-verification/src/issuer-def-resolver.ts
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * An interface for Resolution of issuers for a namespace
+ */
+export interface IssuerDefinitionResolver {
+  /**
+   * @param namespace
+   * @returns IIssuerDefinition for the given namespace
+   */
+  getRoleIssuerDefinition(namespace: string): Promise<IIssuerDefinition>;
```
jrhender, line 14: Maybe shorten it to `getIssuerDefinition`. More concise and more generic (as this code could conceivably be used for credentials other than role credentials) 
```suggestion
  getIssuerDefinition(namespace: string): Promise<IIssuerDefinition>;
```

packages/vc-verification/src/issuer-verification.ts
```diff
@@ -0,0 +1,230 @@
...
+import { CredentialResolver, IssuerDefinitionResolver } from '.';
+import {
+  IVerifiableCredential,
+  VerificationResult,
+  OffChainClaim,
+} from './models';
+import { upgradeChainId } from './upgrade-chainid';
+
+export class IssuerVerification {
+  private _provider: providers.Provider;
```
jrhender, line 15: Doesn't need to be an instance variable as it isn't used, as far as I can tell

packages/vc-verification/src/issuer-verification.ts
```diff
@@ -0,0 +1,230 @@
...
+  private _resolver: Resolver;
+  private _registrySetting: RegistrySettings;
```
jrhender, line 17: Also doesn't need to be an instance variable as it is just used to instantiate `_resolver`

jrhender (approved):
As I'm away for the next few days, I'm approving so that the merge is unblocked and it can potentially be published before I get back. That said, @JGiter or @Harasz if you guys could review to get a second pair of eyes on it before it is merged, I think that would be ideal

packages/vc-verification/src/Ipfs-credential-resolver.ts
```diff
@@ -0,0 +1,81 @@
...
+   * @param serviceEndpoint
+   * @returns {string}
+   */
+  async getCredential(did: string, role: string) {
+    const offChainClaims = await this.offchainClaimsOf(did);
+    for (const claim of offChainClaims) {
+      if (claim.claimType === role) {
+        return claim;
+      }
+    }
```
Harasz, line 38:
```suggestion
    return offChainClaims.find({claimType} => claimType === role);
```

packages/vc-verification/src/issuer-verification.ts
```diff
@@ -0,0 +1,230 @@
...
+  ) {
+    if (credential && verifyCredentialProofCallback) {
+      this.verifyChainOfTrustCallback(
+        credential,
+        verifyCredentialProofCallback
+      );
+    } else if (credential) {
+      this.verifyCredentialChainOfTrust(credential);
+    } else {
+      return 'Insufficient information';
```
Harasz, line 61: Why not to throw an error?

packages/vc-verification/src/issuer-verification.ts
```diff
@@ -0,0 +1,230 @@
...
+   * @param role
+   * @returns
+   */
+  private async isRoleIssuerDID(role: string) {
+    const issuers = await this._issuerDefResolver.getRoleIssuerDefinition(role);
+    if (issuers.did && issuers.did.length > 0) {
+      return true;
+    } else {
+      return false;
+    }
```
Harasz, line 171:
```suggestion
    return issuers.did && issuers.did.length > 0;
```

packages/vc-verification/src/issuer-verification.ts
```diff
@@ -0,0 +1,230 @@
...
+      return false;
+    } else {
+      const isClaimVerified = await this.verifyIssuedToken(
+        offChainClaim.issuedToken
+      );
+      if (typeof isClaimVerified == 'string') {
+        return false;
+      } else {
+        return true;
+      }
```
Harasz, line 227:
```suggestion
      return typeof isClaimVerified !== 'string'
```

packages/vc-verification/src/models/verifiable-credentials.ts
Harasz: we should use those interfaces

packages/vc-verification/src/upgrade-chainid.ts
```diff
@@ -0,0 +1,40 @@
...
+ * @param {string} did
+ * @returns {string} DID address in format "did:" method-name ":" method-specific-id ":" address
+ */
+function upgradeDidWithChainId(did: string): string {
+  const { foundChainInfo } = getDIDChain(did);
+  if (foundChainInfo) return did;
+
+  const didParts = did.split(':');
+
+  const chainName = 'volta';
```
Harasz, line 38: Should it be hardcoded just for `volta`?

packages/vc-verification/test/chain-of-trust-test.ts
```diff
@@ -0,0 +1,454 @@
...
+      await adminOperator.update(adminDid, DIDAttribute.ServicePoint, {
+        type: DIDAttribute.ServicePoint,
+        value: {
+          id: adminDid,
+          type: adminRole,
+          serviceEndpoint: credentialUrl,
+        },
+      });
+
+      console.log('helllllooooo');
```
Harasz, line 450: Hello! :wave:

nichonien: Just for debugging, tests aren't finished yet.

Harasz: I know, I know. I just couldn't resist :smile:

packages/vc-verification/src/upgrade-chainid.ts
nichonien: This code comes from `passport-did-auth`, and AFAIK any `did` with `ewc` if for EWC while with `Volta` or without chainId is for Volta. Ref: https://energyweb.atlassian.net/browse/EDR-88

Harasz: Fair enough

packages/vc-verification/src/models/verifiable-credentials.ts
Harasz: It was intended to be generic and it allows you to pass your own types for `credentialSubject` and I think that's enough to make your implementation on top of it. Rest of the parameters come from the specification.

nichonien: Agreed, I was trying to know your thoughts about the naming convention and structure of the `CredentialSubject`. Since all this would mostly come from `sis-hub` and verification code needs to align with it.

Harasz: `CredentialSubject` requires only `id: string` and that's it

packages/vc-verification/src/Ipfs-credential-resolver.ts
```diff
@@ -0,0 +1,77 @@
...
+  async getCredential(did: string, role: string) {
+    const offChainClaims = await this.offchainClaimsOf(did);
+    return offChainClaims.find(
+      (claim) =>
+        claim.claimType === role || utils.namehash(claim.claimType) === role
+    );
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  private async isOffchainClaim(claim: any): Promise<boolean> {
```
Harasz, line 38:
```suggestion
  private async isOffchainClaim(claim: unknown): claim is OffChainClaim {
```
What do you think of that?

packages/vc-verification/src/issuer-def-resolver.ts
```diff
@@ -0,0 +1,54 @@
...
+   * @returns issuers for the namespace from blockchain contract
+   */
+  async getIssuerDefinition(namespace: string): Promise<IIssuerDefinition> {
+    const issuers: IIssuerDefinition = {};
+    const result = await this._roleDefResolver.issuers(
+      namespace.startsWith('0x') ? namespace : utils.namehash(namespace)
+    );
+    const type = await this._roleDefResolver.issuerType(
+      namespace.startsWith('0x') ? namespace : utils.namehash(namespace)
+    );
```
Harasz, line 45: Could we wrap it with `Promise.all` to increase performance?

packages/vc-verification/src/issuer-def-resolver.ts
```diff
@@ -0,0 +1,54 @@
...
+    if (result.dids.length > 0 || result.role.length > 0) {
+      issuers.did = result.dids;
+      issuers.roleName = result.role;
+      //confirm the type representation TODO
+      issuers.issuerType = type.toString();
+    }
+    return issuers;
```
Harasz, line 52: I think it would be better to return undefined instead od empty object, reason: `Boolean({}) === true` and `Boolean(undefined) === false`

packages/vc-verification/src/issuer-verification.ts
```diff
@@ -0,0 +1,221 @@
...
+      );
+    }
+    if (!offChainClaim) {
+      return false;
+    } else {
+      const isClaimVerified = await this.verifyIssuedToken(
+        offChainClaim.issuedToken
+      );
+      return typeof isClaimVerified === 'string';
+    }
```
Harasz, line 219:
```suggestion
    }
    const isClaimVerified = await this.verifyIssuedToken(
      offChainClaim.issuedToken
    );
    return typeof isClaimVerified === 'string';
```
I think we don't need to nest it in `else` statement. Less indentation == more readable code, imo

test/utils/ipfs-daemon.ts
```diff
@@ -0,0 +1,23 @@
...
+    test: true,
+    ipfsBin,
+    ipfsHttpModule,
+  });
+  return ipfsd.apiAddr;
+}
+
+export async function shutDownIpfsDaemon(): Promise<void> {
+  return ipfsd && ipfsd.stop();
+}
```
Harasz, line 23: new line a the end of file

packages/vc-verification/package.json
```diff
@@ -27,6 +36,33 @@
...
+    "ipfsd-ctl": "3.0.0",
+    "ipfs-http-client": "^43.0.0",
+    "js-sha3": "^0.8.0",
+    "jsonwebtoken": "^8.5.1",
+    "mocha": "7.2.0",
+    "run-with-testrpc": "0.3.1",
+    "truffle": "5.4.29",
+    "ts-node": "9.1.1",
+    "typechain": "7.0.0",
+    "typescript": "^4.5.5"
```
Harasz, line 66: We missing https://github.com/energywebfoundation/shared-configs

nichonien: It is at the repo level https://github.com/energywebfoundation/ew-credentials/blob/17d84c97560bca4028a9341bf60fec5c34fc1e49/package.json#L48, though the version needs to be updated.

packages/vc-verification/src/issuer-verification.ts
```diff
@@ -0,0 +1,221 @@
...
+    } else {
+      throw new Error('Not sufficient information');
+    }
+  }
+
+  /**
+   * Verifies chain of trust for a given holder's DID and role
+   * @param {string} credential
+   */
+  async verifyCredentialChainOfTrust(credential: IVerifiableCredential) {
```
JGiter, line 65: I suggest to implement this as `verifyChainOfTrustCallback(credential, verifyByRoleDefinition)` where `verifyByRoleDefinition` is block staring from line 81

packages/vc-verification/src/issuer-verification.ts
```diff
@@ -0,0 +1,221 @@
...
+  ) {
+    this._issuerDefResolver = issuerDefResolver;
+    this._resolver = new Resolver(provider, registrySetting);
+    this._credentialResolver = credentialResolver;
+  }
+
+  /**
+   * Verifies chain of trust for a holder's credential
+   * @param credential
+   * @param verifyCredentialProofCallback
```
JGiter, line 40:
```suggestion
   * @param verifyCredentialProofCallback verification applied to each issuer in issuers hierarchy. By default issuer is verified according to role definition 
```

packages/vc-verification/src/issuer-def-resolver.ts
nichonien: Do you think `fail-first` reason for using `Promise.all`? IMO the above case should always be consistent in any scenarios, so the performance remains same?

packages/vc-verification/src/Ipfs-credential-resolver.ts
nichonien: This might throw error, reason being the method is async and also the `Object` property doesn't handle unknown type I guess.

packages/vc-verification/src/issuer-verification.ts
nichonien: I think it is better as the current implementation, as the caller could simply pass the `credential` without thinking of the possible `callbacks` maybe? assuming the default mechanism is the RoleDefinition way.

packages/vc-verification/src/issuer-verification.ts
nichonien: Would add more descriptive comment

packages/vc-verification/src/issuer-verification.ts
JGiter: I mean to change only implementation of this method. Also it make sense to name it as `verifyChainOfTrustByRoleDefinitions`

nichonien: I would update the method name but I didn't really understand the implementational changes, could you please elaborate more.

JGiter: My goal is to avoid duplication in `verifyChainOfTrust` and `verifyChainOfTrustCallback`. Ideally `verifyChainOfTrustCallback` would be one line `verifyChainOfTrust(credential, this.verifyByRoleDefinition)`

nichonien: The `verifyChainOfTrustCallback` would accept external libraries for the callback. AFAIK this would lead to general purpose VC Verification something like `did-kit` `verifyCredential` method, thus the code would be different, or still I am getting confused?

packages/vc-verification/package.json
Harasz: fair enough

packages/vc-verification/src/issuer-def-resolver.ts
Harasz: `Promise.all()` will run it at the same moment and not in order. In this case we are blocking thread just once and not twice. 
Second problem here is: we are passing the same argument maybe we can do smth like `const resolvedNamespace = namespace.startsWith('0x') ? namespace : utils.namehash(namespace)` and then pass to those functions `resolvedNamespace`

packages/vc-verification/src/Ipfs-credential-resolver.ts
Harasz: Ohh yeah, type guard doesn't support promises yet :/
```suggestion
  private async isOffchainClaim(claim: unknown): Promise<boolean> {
    if (!claim) return false;
    if (typeof claim !== 'object') return false;
    const offChainClaimProps = [
      'claimType',
      'claimTypeVersion',
      'issuedToken',
      'iss',
    ];
    const claimProps = Object.keys(claim);
    return offChainClaimProps.every((p) => claimProps.includes(p));
  }
```

Harasz: But we can still use `unknown` type and it's highly recommended pattern in typescript projects to avoid `any` and use `unknown` instead

Harasz: https://stackoverflow.com/questions/51439843/unknown-vs-any
https://dev.to/arikaturika/typescript-why-to-use-unknown-instead-of-any-41i8

packages/vc-verification/src/Ipfs-credential-resolver.ts
```diff
@@ -0,0 +1,77 @@
...
+    };
+
+    const filterOutMaliciousClaims = (
+      item: OffChainClaim | undefined
+    ): item is OffChainClaim => {
+      return !!item;
+    };
+
+    const didDocument = await this._resolver.read(did);
+    const services = didDocument.service || [];
```
Harasz, line 64: services has type `any`, maybe we can add type to it, at least `{serviceEndpoint: string}`

packages/vc-verification/src/Ipfs-credential-resolver.ts
nichonien: Yeah, agreed!

packages/vc-verification/src/issuer-verification.ts
JGiter: As discussed will put this aside for now

packages/vc-verification/src/Ipfs-credential-resolver.ts
nichonien: It is string by default, since the `services` are of `IServiceEndpoint` type.

Harasz: Okay, then we can do `const services: IServiceEndpoint[] = didDocument.service || [];`. Because in my vscode I see `any` type and this line is marked as error

packages/vc-verification/src/issuer-verification.ts
```diff
@@ -0,0 +1,216 @@
...
+    }
+  }
+
+  /**
+   * Verifies chain of trust for a given holder's DID and role
+   * @param {string} credential
+   */
+  async verifyChainOfTrustByRoleDefinition(credential: IVerifiableCredential) {
+    let isVcVerified = false;
+    let hasParent = true;
```
JGiter, line 66: Last small suggestion. The return semantic of this method is not consistent. On failed verification method throws, on successful method returns boolean. I think on success we can just return from method

packages/vc-verification/src/Ipfs-credential-resolver.ts
```diff
@@ -0,0 +1,83 @@
...
+   */
+  async getCredential(did: string, role: string) {
+    const offChainClaims = await this.offchainClaimsOf(did);
+    return offChainClaims.find(
+      (claim) =>
+        claim.claimType === role || utils.namehash(claim.claimType) === role
+    );
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
Harasz, line 41:
```suggestion
```

packages/vc-verification/src/issuer-verification.ts
```diff
@@ -0,0 +1,216 @@
...
+        const issuerCredential = await this._credentialResolver.getCredential(
+          credential.issuer,
+          role
+        );
+        // @ts-ignore
+        const result = await verifyCredentialProofCallback(issuerCredential);
+        if (result) {
+          // @ts-ignore
+          credential = issuerCredential;
+        } else {
```
Harasz, line 140: Could you explain why `// @ts-ignore` is needed here?
Second thing, assigning something to `credential` is making side effects for this method, I dont think that is intended. Lets imagine this example 
```
const credential = {...} // some credential
verifyChainOfTrustCallback(credential, cb)
// at this moment credential could have different value
doSmthElse(crednetial) // Could throw an error
```

packages/vc-verification/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "outDir": "./dist",
+    "resolveJsonModule": true
+  },
+  "include": ["src/**/*", "ethers/**/*"]
+}
```
Harasz, line 8: new line at the end of file

typings/ipfsd-ctl/index.d.ts
```diff
@@ -0,0 +1 @@
+declare module 'ipfsd-ctl';
```
Harasz, line 1: This package has built-in type declarations, why do we need to declare module like this? cc: `ipfs`, `ipfs-http-client`

typings/jsonwebtoken/index.d.ts
```diff
@@ -0,0 +1 @@
+declare module 'jsonwebtoken';
```
Harasz, line 1: https://www.npmjs.com/package/@types/jsonwebtoken
Could we use this types?

packages/vc-verification/src/issuer-verification.ts
nichonien: This is an incomplete implementation, mentioned in the description as well. Also it is unclear that what interface or type of object the callback function would take while iterating over the chain of trust. So, just to move ahead with the development it was put there.

typings/ipfsd-ctl/index.d.ts
nichonien: It throws error in the implementation even after successful build and existing in the node_modules. The same case is there in `ew-did-regsitry` as well. https://github.com/energywebfoundation/ew-did-registry/tree/development/typings

typings/jsonwebtoken/index.d.ts
nichonien: Before adding it explicitly to types, I used the above as well. Didn't work for me.

packages/vc-verification/src/upgrade-chainid.ts
```diff
@@ -0,0 +1,40 @@
...
+ */
+export function upgradeChainId(offChainClaim: OffChainClaim) {
+  let invalidDIDProperty = false;
+  let key: keyof OffChainClaim;
+  for (key in offChainClaim) {
+    if (didFormatFields.includes(key)) {
+      const expectedEthrDID = offChainClaim[key] as string;
+      if (isValidErc1056(expectedEthrDID)) {
+        // @ts-ignore
+        offChainClaim[key] = upgradeDidWithChainId(expectedEthrDID);
```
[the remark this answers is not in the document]
nichonien: I get `Type string is not assignable to type never` error. I tried to eliminate it but wasn't successful. If you have any suggestion, please let me know.

packages/vc-verification/src/issuer-verification.ts
Harasz: fair enough

typings/ipfsd-ctl/index.d.ts
Harasz: A little bit weird, maybe we can add a task to jira to fix that or at least try to fix

packages/vc-verification/src/upgrade-chainid.ts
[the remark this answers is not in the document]
Harasz: Probably adding `[x: string]: string` to `OffChainClaim` type should help. But the problem is also with updating parametr that we are passing to method, in this case `offChainClaim`, which is causing side effects for function. 
Read about pure functions in JS, ex. https://blog.greenroots.info/what-are-pure-functions-and-side-effects-in-javascript

typings/ipfsd-ctl/index.d.ts
nichonien: I took some time and tried to fix it, we can have a dedicated task for it maybe later based on the priorities.

JGiter (approved):

PR #14, opened 2022-03-30
use provider instead of signer

jrhender:


nichonien:
@jrhender I have made this changes to my PR #13 . So, closing it.

PR #16, opened 2022-05-02
feat(revocation): add onchain claim revocation code

nichonien:
This PR adds code for revocation of onchain claims / roles. The code haas been added to the `onchain-claims`  package.

### DIFF
--- a/packages/onchain-claims/src/claim-revocation.ts
+++ b/packages/onchain-claims/src/claim-revocation.ts
@@ -1,4 +1,4 @@
-import { Event, utils, providers } from 'ethers';
+import { Event, utils } from 'ethers';
 import { addressOf, EwSigner } from '@ew-did-registry/did-ethr-resolver';
 import {
   ClaimsRevocationRegistry,
@@ -10,7 +10,7 @@ export class ClaimRevocation {
 
   /**
    * @param provider
-   * @param revocationRegistryOnChainAddr - Address of the on chain claim's RevocationRegistry
+   * @param revocationRegistryOnChainAddr - Address of the claim's RevocationRegistry
    */
   constructor(revoker: EwSigner, revocationRegistryOnChainAddr: string) {
     this._revocationRegistry = ClaimsRevocationRegistry__factory.connect(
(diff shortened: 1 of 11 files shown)


JGiter (dismissed):

packages/onchain-claims/src/claim-revocation.ts
```diff
@@ -0,0 +1,130 @@
...
+      const tx = await this._revocationRegistry.revokeClaim(
+        namespaceHash,
+        subjectAddress,
+        revokerAddress
+      );
+      const receipt = await tx.wait();
+      const event = receipt.events?.find((e: Event) => e.event === 'Revoked');
+      if (!event) return false;
+    } catch (error) {
+      if (error instanceof Error) {
```
Harasz, line 49: What was the reason to add such check?

packages/onchain-claims/src/claim-revocation.ts
```diff
@@ -0,0 +1,130 @@
...
+   * @returns Promise<boolean>
+   */
+  async revokeClaimforDIDs(
+    namespace: string,
+    subjects: string[],
+    revoker: string
+  ): Promise<boolean> {
+    const namespaceHash = utils.namehash(namespace);
+    const revokerAddress = addressOf(revoker);
+    const revocationSubjects = [];
```
Harasz, line 72:
```suggestion
    const revocationSubjects = subjects.map(subject => addressOf(subject))
```

Harasz (commented):
I see a lot of eslint warnings and errors

JGiter (approved):

packages/onchain-claims/src/claim-revocation.ts
nichonien: If you are asking about `if (error instanceof Error) `, then it was added when we upgraded `typescript`. For some reason the error regarding the use of `any` was not getting eliminated and it is there for all such call in `ew-did-registry` if I remember correctly.

nichonien: Most of the eslint warnings are for `no-unresolved`. I would put some time and work on it later.

packages/onchain-claims/src/claim-revocation.ts
```diff
@@ -0,0 +1,127 @@
+import { Event, utils, providers } from 'ethers';
```
Harasz, line 1: `[@typescript-eslint/no-unused-vars] 'providers' is defined but never used.`

packages/onchain-claims/src/claim-revocation.ts
Harasz: Okay, why we catch error and then re-throw it?

PR #17, opened 2022-05-10
bug/DID type issuer is not authorized

JGiter:


packages/vc-verification/src/issuer-verification.ts
```diff
@@ -188,7 +188,7 @@ export class IssuerVerification {
     );
     if (issuers && issuers.did && issuers.did.length > 0) {
       for (let i = 0; i < issuers.did.length; i++) {
-        if (issuers.did[i] == addressOf(issuerDID)) {
+        if (issuers.did[i] == issuerDID) {
```
nichonien, line 191: @JGiter The `issuers.did[I]` are addresses that comes from the `RoleResolvers`. So we have to compare the addresses in this scenario. Thus `addressOf` was being used to for IssuerDID.

nichonien: Or maybe we can generalise the role-resolvers to return DIDs instead of addresses?

JGiter: So you think that return type of `IssuerDefinitionResolver.getIssuerDefinition` should be not `IIssuerDefinition`?

nichonien: It should be there, but we should convert those `addresses` into DIDs before adding to the `IIssuerDefinition`. So, in general the issuers from any source ( Cache / ENS / something like VUI issuer interface ) would always return in the form of DIDs.

packages/vc-verification/src/issuer-def-resolver.ts
```diff
@@ -48,9 +50,18 @@ export class EthersProviderIssuerDefinitionResolver
...
     ]);
+    const chainID = (await this._provider.getNetwork()).chainId;
+    let issuerDIDs: string[] = result.dids;
+    if (chainID == 73799) {
+      issuerDIDs = issuerDIDs.map(
+        (issuerDIDs) => `did:ethr:volta:${issuerDIDs}`
+      );
+    } else {
+      issuerDIDs = issuerDIDs.map((issuerDIDs) => `did:ethr:ewc:${issuerDIDs}`);
+    }
```
jrhender, line 61: - I would check explicitly for EWC otherwise there could be confusing results. I think it would be confusing (and technically incorrect) if a non-EWC provider is configured but EWC DIDs are returned.
- Better to import the CHAIN_IDs https://github.com/energywebfoundation/ew-credentials/blob/bug/did_type_issuer_is_not_authorized/packages/credential-governance/src/chain-constants.ts
```suggestion
    if (chainID == VOLTA_CHAIN_ID) {
      issuerDIDs = issuerDIDs.map(
        (issuerDIDs) => `did:ethr:volta:${issuerDIDs}`
      );
    } else if (chainID === EWC_CHAIN_ID) {
      issuerDIDs = issuerDIDs.map((issuerDIDs) => `did:ethr:ewc:${issuerDIDs}`);
    }
    else {
      throw Exception()
    }
```

jrhender (changes requested):
I think the approach makes sense but added some suggestions.

nichonien:
@JGiter I think we can close this? The changes are no more needed.

packages/vc-verification/src/issuer-verification.ts
JGiter: Ok

JGiter:
Closed, because the bug doesn't appear anymore

PR #18, opened 2022-05-16
bug/ICS-204 domains not cached

JGiter:
This fixes one of the reasons of https://energyweb.atlassian.net/browse/ICS-204

Domains will be read from RoleResolverV2 as we not using PublicResolver any more. Or may be configure this depending on chain?

### DIFF
--- a/packages/credential-governance/src/constants.ts
+++ b/packages/credential-governance/src/constants.ts
@@ -2,6 +2,5 @@ import { utils } from 'ethers';
 
 const { parseEther } = utils;
 
-export const emptyAddress = '0x0000000000000000000000000000000000000000';
 export const PRINCIPAL_THRESHOLD = parseEther('100');
 export const WITHDRAW_DELAY = 5;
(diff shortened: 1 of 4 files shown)


Harasz (approved):

jrhender (changes requested):
I don't think this change is necessary.

jrhender:
`getSubdomainsUsingResolver` will use all of the configured resolvers from the logs of the `domainNotifier` contract https://github.com/energywebfoundation/ew-credentials/blob/fdd462e011f182a04b0116b79f05fd6984b8b64e/packages/credential-governance/src/domain-hierarchy.ts#L107
The public resolver is optional.  https://github.com/energywebfoundation/ew-credentials/blob/fdd462e011f182a04b0116b79f05fd6984b8b64e/packages/credential-governance/src/domain-hierarchy.ts#L70
I don't see the value in removing backwards compatibility at this stage, other than cleaning up the code. If this is done, it should be done very intentionally and carefully.

JGiter:
Closed as risk of break existing code is still high

PR #19, opened 2022-05-18
Ec-29 verify chain of trust for offChainClaims

nichonien:
This PR makes changes to accommodate issuance verification for both `OffChainClaims` and `Verifiable Credentials`.

### DIFF
--- a/packages/vc-verification/src/Ipfs-credential-resolver.ts
+++ b/packages/vc-verification/src/Ipfs-credential-resolver.ts
@@ -1,7 +1,7 @@
 import { providers, utils } from 'ethers';
 import { DidStore } from '@ew-did-registry/did-ipfs-store';
 import { IDidStore } from '@ew-did-registry/did-store-interface';
-import { OffChainClaim } from './models';
+import { IVerifiableCredential, OffChainClaim } from './models';
 import { Resolver } from '@ew-did-registry/did-ethr-resolver';
 import {
   RegistrySettings,
@@ -31,6 +31,21 @@ export class IpfsCredentialResolver implements CredentialResolver {
    * @returns
    */
   async getCredential(did: string, role: string) {
+    const credentials = await this.credentialsOf(did);
+    return credentials.find(
+      (claim) =>
+        claim.credentialSubject.role.namespace === role ||
+        utils.namehash(claim.credentialSubject.role.namespace) === role
+    );
+  }
+
+  /**
+   * Fethces credential for the given did and role
+   * @param did
+   * @param role
+   * @returns
+   */
+  async getOffChainClaim(did: string, role: string) {
     const offChainClaims = await this.offchainClaimsOf(did);
     return offChainClaims.find(
       (claim) =>
@@ -78,5 +93,42 @@ export class IpfsCredentialResolver implements CredentialResolver {
       .filter(this.isOffchainClaim)
       .map(transformClaim)
       .filter(filterOutMaliciousClaims);
+  }
+
+  async isCredential(credential: unknown): Promise<boolean> {
+    if (!credential) return false;
+    if (typeof credential !== 'object') return false;
+    const credentialProps = [
+      '@context',
+      'id',
+      'type',
+      'issuer',
+      'issuanceDate',
+      'credentialSubject',
+      'proof',
+    ];
+    const credProps = Object.keys(credential);
+    return credentialProps.every((p) => credProps.includes(p));
+  }
+
+  private async credentialsOf(did: string): Promise<IVerifiableCredential[]> {
+    const filterOutMaliciousCredentials = (
+      item: IVerifiableCredential | undefined
+    ): item is IVerifiableCredential => {
+      return !!item;
+    };
+
+    const didDocument = await this._resolver.read(did);
+    const services: IServiceEndpoint[] = didDocument.service || [];
+    return (
+      await Promise.all(
+        services.map(async ({ serviceEndpoint }) => {
+          const claimToken = await this._ipfsStore.get(serviceEndpoint);
+          return jwt.decode(claimToken) as IVerifiableCredential;
+        })
+      )
+    )
+      .filter(this.isCredential)
+      .filter(filterOutMaliciousCredentials);
   }
 }
--- a/packages/vc-verification/src/Ipfs-credential-resolver.ts
+++ b/packages/vc-verification/src/Ipfs-credential-resolver.ts
@@ -25,22 +25,22 @@ export class IpfsCredentialResolver implements CredentialResolver {
   }
 
   /**
-   * Fethces credential for the given did and role
+   * Fetches credential for the given did and role for a vc issuance hierarchy
    * @param did
-   * @param role
+   * @param namespace
    * @returns
    */
-  async getCredential(did: string, role: string) {
+  async getCredential(did: string, namespace: string) {
     const credentials = await this.credentialsOf(did);
     return credentials.find(
       (claim) =>
-        claim.credentialSubject.role.namespace === role ||
-        utils.namehash(claim.credentialSubject.role.namespace) === role
+        claim.credentialSubject.role.namespace === namespace ||
+        utils.namehash(claim.credentialSubject.role.namespace) === namespace
     );
   }
 
   /**
-   * Fethces credential for the given did and role
+   * Fetches issued token for the given did and role for an OffChainClaim issuance hierarchy
    * @param did
    * @param role
    * @returns
@@ -99,9 +99,11 @@ export class IpfsCredentialResolver implements CredentialResolver {
       .filter(filterOutMaliciousClaims);
   }
 
-  async isCredential(credential: unknown): Promise<boolean> {
-    if (!credential) return false;
-    if (typeof credential !== 'object') return false;
+  isVerifiableCredential(
+    vc: IVerifiableCredential | unknown
+  ): vc is IVerifiableCredential {
+    if (!vc) return false;
+    if (typeof vc !== 'object') return false;
     const credentialProps = [
       '@context',
       'id',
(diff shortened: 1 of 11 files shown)


packages/vc-verification/src/issuer-verification-offchainclaims.ts
```diff
@@ -0,0 +1,134 @@
...
+  async verifyIssuerAuthority(
+    namespace: string,
+    issuerDID: string
+  ): Promise<boolean> {
+    const issuers = await this._issuerDefResolver.getIssuerDefinition(
+      namespace
+    );
+    if (issuers && issuers.did && issuers.did.length > 0) {
+      for (let i = 0; i < issuers.did.length; i++) {
+        if (issuers.did[i] == addressOf(issuerDID)) {
```
JGiter, line 114: @nichonien Currently this check is always returns false. So we need either to change issuerDID to issuerAddr or (which makes more sense to me) removed addressOf
```suggestion
        if (issuers.did[i] === addressOf(issuerDID)) {
```

packages/vc-verification/src/issuer-verification-offchainclaims.ts
```diff
@@ -0,0 +1,134 @@
...
+        return true;
+      } else if (roleIssuers && roleIssuers.roleName) {
+        const offChainClaim = await this._credentialResolver.getOffChainClaim(
+          currentIssuerDID,
+          roleIssuers.roleName
+        );
+        if (!offChainClaim) {
+          throw new Error('No credential found');
+        } else {
+          let nextIssuerDID;
```
jrhender, line 55: The `else` is not needed because of the `Error` thrown in the `if`. This removes some of the nesting
```suggestion
        }
        let nextIssuerDID;
```

packages/vc-verification/src/issuer-verification-offchainclaims.ts
```diff
@@ -0,0 +1,134 @@
...
+      if (!(await this.verifyIssuerAuthority(role, currentIssuerDID))) {
+        throw new Error('Issuer is not allowed to issue credential');
+      }
+      const roleIssuers = await this._issuerDefResolver.getIssuerDefinition(
+        role
+      );
+      if (await this.isRoleIssuerDID(role)) {
+        return true;
+      } else if (roleIssuers && roleIssuers.roleName) {
+        const offChainClaim = await this._credentialResolver.getOffChainClaim(
```
jrhender, line 48: IMO, it's better to return the `issuedToken` from `getOffChainClaim`, IMO. This will avoid the `offChainClaim.issuedToken` references on line 56 and 58 (which reduces the knowledge this method needs on `offChainClaim`'s internals) and could potential remove the need for the additional `if` on line 56
```suggestion
        const issuedToken = await this._credentialResolver.getOffChainClaim(
```

packages/vc-verification/src/issuer-verification-offchainclaims.ts
```diff
@@ -0,0 +1,134 @@
...
+        } else {
+          let nextIssuerDID;
+          if (offChainClaim.issuedToken) {
+            nextIssuerDID = await this.verifyIssuedToken(
+              offChainClaim.issuedToken
+            );
+          }
+          if (nextIssuerDID) {
+            currentIssuerDID = nextIssuerDID;
+            if (roleIssuers && roleIssuers.roleName) {
```
jrhender, line 63: I don't think this `if` is required because it is already performed on line 47

packages/vc-verification/src/issuer-verification-offchainclaims.ts
```diff
@@ -0,0 +1,134 @@
...
+              role = roleIssuers.roleName;
+            }
+          } else {
+            throw new Error('The credential is invalid');
+          }
```
jrhender, line 68: I would consider inverting this `if` to avoid the nesting (similar to how you've done on line 52)

packages/vc-verification/src/issuer-verification-offchainclaims.ts
```diff
@@ -0,0 +1,134 @@
...
+      );
+      if (await this.isRoleIssuerDID(role)) {
+        return true;
+      } else if (roleIssuers && roleIssuers.roleName) {
+        const offChainClaim = await this._credentialResolver.getOffChainClaim(
+          currentIssuerDID,
+          roleIssuers.roleName
+        );
+        if (!offChainClaim) {
+          throw new Error('No credential found');
```
jrhender, line 53:
```suggestion
          throw new Error('Unable to resolve the issuer credential to verify their authority');
```

jrhender (changes requested):
Some suggestions for improving the core `chainOfTrust` function

packages/vc-verification/src/issuer-verification-offchainclaims.ts
nichonien: This should be addressed in https://energyweb.atlassian.net/browse/EC-30

JGiter (approved):

jrhender:
@nichonien Is it expected that that the 3 github jobs are failing?

nichonien:
> @nichonien Is it expected that that the 3 github jobs are failing?

No, all three jobs are failing because of `eslint` config which I need to change (move the config to shared-config and make it optional for other repos). Would do it once the priority tasks are done.

packages/vc-verification/src/Ipfs-credential-resolver.ts
```diff
@@ -31,11 +31,30 @@ export class IpfsCredentialResolver implements CredentialResolver {
...
+    const credentials = await this.credentialsOf(did);
+    return credentials.find(
+      (claim) =>
+        claim.credentialSubject.role.namespace === role ||
+        utils.namehash(claim.credentialSubject.role.namespace) === role
+    );
+  }
+
+  /**
+   * Fethces credential for the given did and role
```
jrhender, line 43: The documentation for `getClaimIssuedToken` and `getCredential` are the same which seems a bit confusing as they are different methods. Could be a bit more specific in the documentation please?

packages/vc-verification/src/Ipfs-credential-resolver.ts
```diff
@@ -79,4 +98,43 @@ export class IpfsCredentialResolver implements CredentialResolver {
       .map(transformClaim)
       .filter(filterOutMaliciousClaims);
   }
+
+  async isCredential(credential: unknown): Promise<boolean> {
```
jrhender, line 102: IMO, using `VerifiableCredential` or `vc` is better. I think that an `offChainClaim` is technically a "credential" but it isn't a "verifiable" credential
```suggestion
  async isVerifiableCredential(vc: unknown): Promise<boolean> {
```

packages/vc-verification/src/Ipfs-credential-resolver.ts
```diff
@@ -31,11 +31,30 @@ export class IpfsCredentialResolver implements CredentialResolver {
    * @returns
    */
   async getCredential(did: string, role: string) {
+    const credentials = await this.credentialsOf(did);
+    return credentials.find(
+      (claim) =>
+        claim.credentialSubject.role.namespace === role ||
+        utils.namehash(claim.credentialSubject.role.namespace) === role
```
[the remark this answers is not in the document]
nichonien: This is for the case when you fetch the `RoleDefinition` and the issuer type is `Role`. This role is namehash (since it is coming from ens contract). This namehash(role) becomes the role (to be queried) for the verification of next issuer. Thus in the next step the `namehash(role)` from ens and issuer `did` is the new params for the issuer verification.

nichonien:
> @nichonien Some general comments
> 
> * It seems to me like there is some duplicated code between `issuer-verification-vc` and issuer-verification-offchainclaim`? Would you agree with this? If so, how long do you think it would take to reduce this code duplication? Maybe if it takes less then a half-day to do, then it is worth doing.
> * On a related note, I think we should note that the entire hierarchy needs to be composed of VCs or offChainClaims for the verification to work currently, right? (There can't be a mix in the hierarchy). I think this is okay, but we should note this as a limitation in the class/method documentation

Reg 1 - The reason behind the duplication of code is that when we deprecate the `OffChainClaims` issuance hierarchy verification, we don't have to change any piece of code in `VC` issuance hierarchy verification. Keeping a common code would make it necessarily to make changes(we had discussed this in the call).

Reg 2- Yes, I totally agree with you. Had discussed the same with Dmitry initially. Would document it within the class for now.

jrhender (dismissed):
Looks good!

JGiter (changes requested):
`verifiyChainOfTrustClaims` Is somewhat difficult for me to read. What do you think of creating template method in base issuer-verification:
```
async verifyChainOfTrustClaims(issuerDID: string, role: string) {
   const {issuerType, did, roleName} = await this._issuerDefResolver.getIssuerDefinition(
        role
      );
   switch (issuerType) {
      case 'DID':
         return did?.includes(issuerDID);
      case 'ROLE': {
         const credential = await this.getCredential(issuerDID, roleName);
         if (!credential) return false;
         return verifyChainOfTrust(credential.issuer, credential.credentialSubject.role.namespace)
      }
   }
```
two subclasses overrides getCredential() and both returns IVerifiableCredential. I propose same return type since we are moving to vc standard, but this is not necessary and case 'ROLE' can be different

nichonien:
> `verifiyChainOfTrustClaims` Is somewhat difficult for me to read. What do you think of creating template method in base issuer-verification:
> 
> ```
> async verifyChainOfTrustClaims(issuerDID: string, role: string) {
>    const {issuerType, did, roleName} = await this._issuerDefResolver.getIssuerDefinition(
>         role
>       );
>    switch (issuerType) {
>       case 'DID':
>          return did?.includes(issuerDID);
>       case 'ROLE': {
>          const credential = await this.getCredential(issuerDID, roleName);
>          if (!credential) return false;
>          return verifyChainOfTrust(credential.issuer, credential.credentialSubject.role.namespace)
>       }
>    }
> ```
> 
> two subclasses overrides getCredential() and both returns IVerifiableCredential. I propose same return type since we are moving to vc standard, but this is not necessary and case 'ROLE' can be different

@JGiter Anyhow we will deprecate the `OffChainClaims` verification in near future. I guess it is fine to have a direct implementation rather than having a base implementation followed by specific implementation, since it will always be a Verifiable Credential on long run.

JGiter (approved):

PR #20, opened 2022-05-20
bug/ICL-247 different domains from resolver and from registry

JGiter:
…to registry


JGiter:
Obsoleted in favor of https://github.com/energywebfoundation/ew-credentials/pull/47

PR #21, opened 2022-05-23
EC-34 subject has role before revocation

nichonien:
This PR contains changes to add check for the subject to have a claim/role before revocation.

### DIFF
--- a/packages/onchain-claims/contracts/ClaimsRevocationRegistry.sol
+++ b/packages/onchain-claims/contracts/ClaimsRevocationRegistry.sol
@@ -36,6 +36,8 @@ contract ClaimsRevocationRegistry {
     }
 
     function revokeClaim(bytes32 role, address subject, address revoker) public {
+        ClaimManager cm = ClaimManager(claimManager);
+        require(cm.hasRole(subject, role, 0), "The claim has not been issued");
         require(revokedClaims[role][subject].revokedTimestamp == 0, "The claim is already revoked");
         verifyRevoker(role, revoker);
         revokedClaims[role][subject].revoker = revoker;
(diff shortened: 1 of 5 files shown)


Harasz (approved):

packages/onchain-claims/contracts/ClaimsRevocationRegistry.sol
```diff
@@ -36,6 +36,8 @@ contract ClaimsRevocationRegistry {
     }
 
     function revokeClaim(bytes32 role, address subject, address revoker) public {
+        ClaimManager cm = ClaimManager(claimManager);
+        require(cm.hasRole(subject, role, 0), "The claim has not been issued");
         require(revokedClaims[role][subject].revokedTimestamp == 0, "The claim is already revoked");
```
JGiter, line 41: Not related to this PR, but 
 - why claim is revoked when its timestamp is 0? Revoked claim can have any timestamp
 - `revokeClaimsInList` begs for reusing `revokeClaim`. Why to duplicate this?

nichonien: Regarding 1 - if the `revokedTimestamp` == 0, it means it is not revoked because if it was then it would definitely be some specific time. Thus, the condition tells, the `revokedTimestamp` should be 0 for revocation to happen.

Reg 2 - The reason for having a separate method for bulk revocation is that, it will verify the revoker only once and thus saving cost and time. Also, the number of blockchain interaction is reduced with single call even for 1000s of subjects revocation.

JGiter: Thanks. Got it 1. For the 2 though I would move everything except of verifyRevoker and creating cm in separate method

JGiter (approved):

PR #29, opened 2022-06-14
chore: add and export VC types from credential-governance package

whitneypurdum:
- Adds and exports RoleCredentialSubject, IssuerFields interfaces from credential governance types

### DIFF
--- a/packages/credential-governance/package.json
+++ b/packages/credential-governance/package.json
@@ -62,6 +62,7 @@
     "@ethersproject/abi": "^5.4.0",
     "@ethersproject/abstract-signer": "^5.4.1",
     "@ew-did-registry/proxyidentity": "0.6.2",
+    "@ew-did-registry/credentials-interface": "0.6.3-alpha.527.0",
     "@openzeppelin/contracts": "4.3.3",
     "@openzeppelin/contracts-upgradeable": "4.3.3",
     "@openzeppelin/truffle-upgrades": "1.12.0",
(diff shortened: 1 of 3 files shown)


whitneypurdum:
@nichonien I think ew-credentials is installing a later version of @sphereon pex, hence the type building errors - should I just put the correct(but older) version of sphereon pex that dependencies are using in package.json?

nichonien:
@whitneypurdum looks like the tests are failing after the package-lock update, can you take a look and try to resolve it. Or maybe revert the last two commits for now.

whitneypurdum:
> 

@nichonien Sure thing, I reverted the last 3 commits, so it's just bumping the ew-did version in credential-governance package, but the ci is failing in tests. This was solved by bumping the "@ew-did-registry/credentials-interface" version in the other package.json as well, as the hoisted version was different than the one in credential-governance, and then I ran reset. How would you handle this? It's for sure a node modules issue as I haven't messed with the code at all other than adding and exporting the types.

jrhender (commented):
Why so many additions to the `bundleDependencies` in `package-lock.json`? It seems to me that this is unrelated to the objectives of this PR?

whitneypurdum:
@jrhender the build kept failing because the ci in the lint source code was saying the package-locks were out of sync. I was thinking maybe it was because there was another version of ew-did credentials in another package, so I bumped that and ran npm run reset, which updated the package-lock. This commit has been reverted though, so it isn't in the PR anymore. Still having the issue of ci not working because of package-locks being out of sycn though. Surely though because I bumped this version there should be changes to package-lock after I run install, build, setup at the root level, but there aren't any changes in the package-lock.

PR #30, opened 2022-06-15
chore: add and export RoleCredentialSubject interface

whitneypurdum:
- Add and export interface for RoleCredentialSubject interface


whitneypurdum:
@jrhender @nichonien This is a new PR off master. The only things that were done were adding/exporting the interface, updating the "@ew-did-registry/credentials-interface" version in credential-governance package.json, and then doing an npm install, setup, build at the root level to make sure there were no build errors. I didn't add any other packages, or run reset etc.

whitneypurdum:
> > I still don't really understand the additions to the `bundleDependencies`. Does it make sense to you @nichonien @whitneypurdum ?
> 
> No, additions to the `bundleDependencies` should not happen as such. What version of `node` are you using @whitneypurdum ?

15.10.0

nichonien:
> > No, additions to the `bundleDependencies` should not happen as such. What version of `node` are you using @whitneypurdum ?
> 
> 15.10.0

We are using 16.10 majorly for all the repos. Not sure this would impact but the second answer in tell that it could be the case.

whitneypurdum:
> > > No, additions to the `bundleDependencies` should not happen as such. What version of `node` are you using @whitneypurdum ?
> > 
> > 
> > 15.10.0
> 
> We are using 16.10 majorly for all the repos. Not sure this would impact but the second answer in https://stackoverflow.com/questions/69984660/npm-ci-can-only-install-packages-with-an-existing-package-lock-json-or-npm-shrin tell that it could be the case.

Sure, my bad. I just removed package-lock at root level and re-installed with node 16.10. 

> > > No, additions to the `bundleDependencies` should not happen as such. What version of `node` are you using @whitneypurdum ?
> > 
> > 
> > 15.10.0
> 
> We are using 16.10 majorly for all the repos. Not sure this would impact but the second answer in https://stackoverflow.com/questions/69984660/npm-ci-can-only-install-packages-with-an-existing-package-lock-json-or-npm-shrin tell that it could be the case.

Sure, that was an bad oversight on my part. I removed the package-lock and re-installed using 16.10.

PR #34, opened 2022-06-16
fix: update Issuer Verification to use

whitneypurdum:
- Update Issuer Verification methods to use VerifiableCredential interface


packages/vc-verification/src/credential-resolver.ts
```diff
@@ -1,4 +1,5 @@
-import { IVerifiableCredential } from './models';
```
jrhender, line 1: If we're removing `IVerifiableCredential` usage from the `ew-credentials` can we remove the actual file? https://github.com/energywebfoundation/ew-credentials/blob/31a697c8f8603b7995794b962cdb1dacccd4bb67/packages/vc-verification/src/models/verifiable-credentials.ts#L1 

It looks like this might be a breaking change (to our exports) though so maybe it would make sense to also export `VerifiableCredential` so that clients don't also have to install `'@ew-did-registry/credentials-interface'`

What are your thoughts here?

whitneypurdum: @jrhender sure, so import @ew-did-registry/credentials-interface and export it from this package? is that right?

jrhender: Yes, that's what makes sense to me

whitneypurdum:
It's weird to be that you can run npm run set up, build at root and there are no type/build errors, but there are build errors when the build-and-test happens? Does the build command at root actually run build in sub-packages?
 "build": "patch-package && rm -rf build docs packages/*/dist && lerna run compile"

packages/vc-verification/src/credential-resolver.ts
jrhender: Thanks for the further changes @whitneypurdum . I think it still make sense to remove `IVerifiableCredential` as it seems like it is not being used anymore and is effectively replaced by `VerifiableCredential`

whitneypurdum: @jrhender Ah yes, sorry I forgot to do that. Do you think I should check if its being used in SSI hub or elsewhere? I dont think ICL is using it anymore.

jrhender: I would just remove it IMO. A breaking change is fine at this point

whitneypurdum: @jrhender yes it’s already removed

jrhender: @whitneypurdum are the other interfaces like `CredentialSubject`, `Role`, `IssuerFields` and `Proof` still needed or can those be removed too now that `IVerifiableCredential` is removed?

jrhender (approved):
Looks good, thanks!

PR #35, opened 2022-06-21
feat: add `defaultExpiration` property to role definition

Harasz:
Add `defaultExpiration` property to role definition

### DIFF
--- a/packages/credential-governance/src/types/domain-definitions.ts
+++ b/packages/credential-governance/src/types/domain-definitions.ts
@@ -25,6 +25,8 @@ export interface IRoleDefinitionV2 extends IRoleDefinitionText {
   issuer: IIssuerDefinition;
   revoker: IRevokerDefinition;
   enrolmentPreconditions: { type: PreconditionType; conditions: string[] }[];
+  // Default seconds value after which the role is considered expired, if empty then no expiry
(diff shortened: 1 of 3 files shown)


packages/credential-governance/src/types/domain-definitions.ts
```diff
@@ -25,6 +25,8 @@ export interface IRoleDefinitionV2 extends IRoleDefinitionText {
   issuer: IIssuerDefinition;
   revoker: IRevokerDefinition;
   enrolmentPreconditions: { type: PreconditionType; conditions: string[] }[];
+  // Default seconds value after which the role is considered expired, if empty then no expiry
+  defaultExpiration?: number;
```
jrhender, line 29: To me `expiration` implies a precise point in time while `validityPeriod` is a period from which an `expiration` can be calculated during issuance
```suggestion
  defaultValidityPeriod?: number;
```

jrhender (commented):
I think that might be the only change. However, could you please add a test or adjust the existing test cases to consider `defaultExpiration`? 
Probably we should test for both the case of a roles with the `defaultExpiration` and roles with the `defaultExpiration`

jrhender (approved):
LGTM

PR #37, opened 2022-06-23
task/EC-35 verification of revocation

JGiter:


packages/vc-verification/src/revocation-verification.ts
```diff
@@ -26,77 +44,101 @@ export class RevocationVerification {
...
   }
 
   /**
-   * Verifies revoker's authority
-   * @param namespace for which the revoker's authority is to be verified
-   * @param revokerDID whose authority needs to be verified
-   * @returns
+   * Verifies that revoker is authorized to revoke role credential
+   * @param role fully qualified role name
+   * @param revoker DID of revoker
```
nichonien, line 81:
```suggestion
   * @param revoker DID of revoker
   * @param role fully qualified role name
```

packages/vc-verification/src/revocation-verification.ts
```diff
@@ -26,77 +44,101 @@ export class RevocationVerification {
...
 
   /**
-   * Checks if the purpose of the issued credential is revocation
-   * @param purpose statusPurpose from StatusListCredential
-   * @returns
+   * Verifies that role credential was issued to subject by authorized issuer
+   * @param subject DID of revoker
+   * @param role fully qualified role name
+   *
+   * @todo reuse in VCIssuerVerification
```
nichonien, line 110: Maybe we can move this to `VCIssuerVerification` and reuse it here. This way we won't need to initialise `VCIssuerVerification` at all [here](https://github.com/energywebfoundation/ew-credentials/blob/a1b7cb3e16a8416f3bf7c046e3151d2bfec41e26/packages/vc-verification/src/revocation-verification.ts#L40)

nichonien (approved):
Thanks @JGiter !

packages/vc-verification/src/errors.ts
```diff
@@ -0,0 +1,62 @@
+import type { IRoleDefinitionV2 } from '@energyweb/credential-governance';
+
+export class NoCredential extends Error {
```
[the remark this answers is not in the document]
JGiter: I don't see what it gives. I can't imagine what `NoCredential` can be confused with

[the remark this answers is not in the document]
Harasz: I am just saying, it's a common pattern for errors and exceptions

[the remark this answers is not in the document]
JGiter: May be following this pattern has some benefits, but it is unclear to me. But in any case I think we need to discuss with team if we are going to apply this pattern to our stack. By the could you share reference to this pattern?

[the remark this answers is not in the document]
Harasz: It's not a architecture pattern. We are using it in other places, at least in most of places. NestJs is also fallowing this rule, they are adding `Exception` suffix to all http error classes

nichonien (dismissed):

packages/vc-verification/src/credential-verification.ts
```diff
@@ -0,0 +1,58 @@
...
+  constructor(
+    private credentialResolver: CredentialResolver,
+    private issuerVerification: VCIssuerVerification
+  ) {}
+
+  /**
+   * Verifies that role credential was issued to subject by authorized issuer
+   * @param subject DID of revoker
+   * @param role fully qualified role name
+   *
```
Harasz, line 23:
```suggestion
   * @return ...
```
?

packages/vc-verification/src/issuer-verification-vc.ts
```diff
@@ -113,7 +114,9 @@ export class VCIssuerVerification {
           credential.issuer as string,
           role
         );
-
+        if (!issuerCredential) {
+          throw new NoCredential(role, credential.issuer as string);
```
Harasz, line 118:
```suggestion
          throw new NoCredential(role, issuerDID(credential.issuer));
```

packages/vc-verification/src/revocation-verification.ts
```diff
@@ -1,102 +1,91 @@
...
+    const { issuer: revoker } = statusList;
+    const credential = await this.credentialVerification.verifyEnrolment(
+      subject,
+      role
+    );
+    if (!credential.credentialStatus) {
+      throw new NoCredentialStatus(role);
+    }
+
+    const statusVerifier = new StatusListEntryVerification(verifyCredential);
```
Harasz, line 48: Maybe we can make it as a private property?

packages/vc-verification/src/revocation-verification.ts
```diff
@@ -1,102 +1,91 @@
...
+    try {
+      await statusVerifier.verifyCredentialStatus(credential.credentialStatus);
+      throw new CredentialNotRevoked(role, subject);
```
Harasz, line 51: why?

JGiter: Verification of revocation assumes that credential was revoked. So non revoked credential means failed verification. May be I am wrong. @jrhender could you confirm this?

jrhender (commented):
Look good, I left a couple comments and I think @Harasz 's comment about the exception thrown after the verification needs to be addressed

packages/vc-verification/src/revocation-verification.ts
jrhender: @JGiter Hmm, I think we don't need to have `await this.statusVerifier.verifyCredentialStatus` here.
I think the verification of a revocation of a credential at a higher level (e.g. in ICL) would be something like:
```
const vc: VerifiableCredential<EWFRoleSubject>;
const { isMarkedAsRevoked, statusListCredential } = StatusListEntryVerification.verifyCredentialStatus(vc.credentialStatus);
let revokedByAuthorizedRevoker = false;
if (isMarkedAsRevoked) {
  revokedByAuthorizedRevoker = RevocationVerification.verifyRevocation(statusListCredential);
}
```
So I don't think we need to check the if the original credential is revoked in `RevocationVerification.verifyRevocation` function? Is this how you see it as well @Harasz @nichonien ?

packages/vc-verification/src/errors.ts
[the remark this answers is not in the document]
JGiter: Thanks. I prefer path without errors :smile:

packages/vc-verification/src/issuer-verification-vc.ts
```diff
@@ -143,46 +135,65 @@ export class VCIssuerVerification {
...
+  }
+
+  /**
+   * Verifies that role credential was issued to subject by authorized issuer
+   * @param subject DID of revoker
+   * @param role fully qualified role credential name
+   *
+   * @returns verified role credential
+   */
+  async verifyCredential(
```
jrhender, line 178: @JGiter I like what you've done here with recursive approach of `verifyCredential` calling `verifyIssuer`, however, I think that it has introduces some duplication with `verifyChainOfTrust` as `verifyChainOfTrust` and `verifyIssuer` essentially do the same thing of "Verifying that role credential was issued to subject by authorized issuer".

IMO, the `verifyCredential`/`verifyIssuer` recursion approach is easier to read than the `verifyChainOfTrust` code, so we should remove the `verifyChainOfTrust`/`verifyChainOfTrustByRoleDefinition`/`verifyChainOfTrustCallback` code.

How does this sound @JGiter @nichonien ?

packages/vc-verification/src/revocation-verification.ts
nichonien: The way I see it is that the `statusVerifier.verifyCredentialStatus` should check the revocation status for the given `StatusListEntry` and the `RevocationVerification.verifyRevocation` verifies the revoker's authority and internally calls the `statusVerifier.verifyCredentialStatus`. So ICL probably has to just rely on `RevocationVerification` from `ew-credentials` for complete verification of `StatusListEntry` and revocation verification. This would reduce calls in ICL and would be more accurate (making a single call for complete verification).

packages/vc-verification/src/issuer-verification-vc.ts
nichonien: It does seem like a good approach but there are some flaws:

1. I would want to have two separate function `verifyCredential` and `verifyIssuer` for verifying the credential and issuer respectively. Since, `ICL` and `SSI-Hub` may need such functionality for individual calls.
2. By definition `verifyCredential` should just verify the credential and probably it could take a `callbackfunction` if needed for flexibility / choice of verification.
3. I would keep `verifyChainOfTrustByRoleDefinition ` and `verifyChainOfTrustCallback ` to verify entire hierarchy of issuance based on the presented credential.
4. Initially the `verifyChainOfTrust` was supposed to be the entry point for chain-of-trust verification based on the passed parameters, but I guess it is not needed now, we can directly call the specific functions or maybe mark them private and export only `verifyChainOfTrust`. 
5. I agree the naming convention could be improved and some code refactoring as well.

JGiter: Yes, we can remove verifyChainOfTrust. It is not used anywhere and its logic is implemented in `verifiy..` methods

JGiter: We can also do the same in issuer-verification-offchain

jrhender (approved):

PR #38, opened 2022-06-30
task/EC-48 align issuer verification API

JGiter:


jrhender (commented):
Looks a good @JGiter , thanks. Even merging as is would be beneficial I think so that it could be integrated into ICL with the `verifyIssuer` API

jrhender (approved):

nichonien (approved):
Thanks @JGiter !

PR #43, opened 2022-07-08
feat: single api for issuer verification

nichonien:
This PR consists of changes:

1. Resolves VCs from IPFS where VC was signed by issuer before publishing.
2. `IssuerVerification` which provides single API to verify issuer authority.

### DIFF
--- a/packages/vc-verification/src/Ipfs-credential-resolver.ts
+++ b/packages/vc-verification/src/Ipfs-credential-resolver.ts
@@ -143,10 +143,11 @@ export class IpfsCredentialResolver implements CredentialResolver {
         services.map(async ({ serviceEndpoint }) => {
           const claimToken = await this._ipfsStore.get(serviceEndpoint);
           let offChainClaim;
+          // expect that JWT has 3 dot-separated parts
           if (claimToken.split('.').length === 3) {
             offChainClaim = jwt.decode(claimToken);
           }
-          return jwt.decode(claimToken) as OffChainClaim;
+          return offChainClaim as OffChainClaim;
         })
       )
     )
@@ -183,9 +184,9 @@ export class IpfsCredentialResolver implements CredentialResolver {
         services.map(async ({ serviceEndpoint }) => {
           const credential = await this._ipfsStore.get(serviceEndpoint);
           let vc;
+          // expect that JWT would have 3 dot-separated parts, VC is non-JWT credential
           if (!(credential.split('.').length === 3)) {
             vc = JSON.parse(credential);
-            delete vc.iat;
           }
           return vc as VerifiableCredential<RoleCredentialSubject>;
         })
--- a/packages/vc-verification/src/Ipfs-credential-resolver.ts
+++ b/packages/vc-verification/src/Ipfs-credential-resolver.ts
@@ -142,6 +142,10 @@ export class IpfsCredentialResolver implements CredentialResolver {
       await Promise.all(
         services.map(async ({ serviceEndpoint }) => {
           const claimToken = await this._ipfsStore.get(serviceEndpoint);
+          let offChainClaim;
+          if (claimToken.split('.').length === 3) {
+            offChainClaim = jwt.decode(claimToken);
+          }
           return jwt.decode(claimToken) as OffChainClaim;
         })
       )
@@ -177,9 +181,12 @@ export class IpfsCredentialResolver implements CredentialResolver {
     return (
       await Promise.all(
         services.map(async ({ serviceEndpoint }) => {
-          const credentialToken = await this._ipfsStore.get(serviceEndpoint);
-          const vc = jwt.decode(credentialToken);
-          delete vc.iat;
+          const credential = await this._ipfsStore.get(serviceEndpoint);
+          let vc;
+          if (!(credential.split('.').length === 3)) {
+            vc = JSON.parse(credential);
+            delete vc.iat;
+          }
           return vc as VerifiableCredential<RoleCredentialSubject>;
         })
       )
--- a/packages/vc-verification/src/Ipfs-credential-resolver.ts
+++ b/packages/vc-verification/src/Ipfs-credential-resolver.ts
@@ -36,7 +36,7 @@ export class IpfsCredentialResolver implements CredentialResolver {
    *  didStore );
    * const credential = credentialResolver.getCredential('did:ethr:1234', 'sampleRole');
    * ```
-   * 
+   *
    * @param did subject DID for which the credential needs to be fetched
    * @param namespace role for which the credential needs to be fetched
    * @returns
@@ -177,8 +177,8 @@ export class IpfsCredentialResolver implements CredentialResolver {
     return (
       await Promise.all(
         services.map(async ({ serviceEndpoint }) => {
-          const credential = await this._ipfsStore.get(serviceEndpoint);
-          const vc = JSON.parse(credential);
+          const credentialToken = await this._ipfsStore.get(serviceEndpoint);
+          const vc = jwt.decode(credentialToken);
           delete vc.iat;
           return vc as VerifiableCredential<RoleCredentialSubject>;
         })
--- a/packages/vc-verification/src/claim-issuer-verification.ts
+++ b/packages/vc-verification/src/claim-issuer-verification.ts
@@ -1,7 +1,7 @@
 import * as jwt from 'jsonwebtoken';
 import { providers } from 'ethers';
 import { ProofVerifier } from '@ew-did-registry/claims';
-import { Resolver } from '@ew-did-registry/did-ethr-resolver';
+import { addressOf, Resolver } from '@ew-did-registry/did-ethr-resolver';
 import { RegistrySettings } from '@ew-did-registry/did-resolver-interface';
 import { CredentialResolver, IssuerResolver } from '.';
 import { OffChainClaim } from './models';
@@ -36,7 +36,7 @@ export class ClaimIssuerVerification {
 
   /**
    * Verifies that `issuer` is authorized to issue `role` claim
-   * 
+   *
    * ```typescript
    * const issuerVerification = new ClaimIssuerVerification(
    * provider,
@@ -127,11 +127,11 @@ export class ClaimIssuerVerification {
       namespace
     );
     if (issuers && issuers.did && issuers.issuerType === 'DID') {
-      for (let i = 0; i < issuers.did.length; i++) {
-        if (issuers.did[i] === issuerDID) {
-          return true;
-        }
-      }
+      return issuers?.did?.find(
+        (d) => addressOf(d).toUpperCase() === addressOf(issuerDID).toUpperCase()
+      )
+        ? true
+        : false;
     }
     let claim;
     if (issuers && issuers.roleName) {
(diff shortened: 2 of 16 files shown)


packages/vc-verification/src/issuer-verification.ts
```diff
@@ -0,0 +1,78 @@
...
+ * A class to provide verification of issuer authority for either VC or OffChainClaim
+ */
+export class IssuerVerification {
+  private credentialResolver: CredentialResolver;
+  private vcIssuerVerification: VCIssuerVerification;
+  private claimIssuerverification: ClaimIssuerVerification;
+
+  constructor(
+    private issuerResolver: IssuerResolver,
+    credentialResolver: CredentialResolver,
```
JGiter, line 23:
```suggestion
    private credentialResolver: CredentialResolver,
```

JGiter: Lets use constructor shorthands

packages/vc-verification/src/issuer-verification.ts
```diff
@@ -0,0 +1,78 @@
...
+    this.vcIssuerVerification = vcIssuerVerification;
+    this.claimIssuerverification = claimIssuerverification;
+  }
+
+  /**
+   * Verifies issuer authority with either Verifiable Credential or OffChainClaim
+   *
+   * ```typescript
+   * const issuerVerification = new IssuerVerification(
+   * credentialResolver,
```
JGiter, line 37:
```suggestion
   * issuerResolver,
   * credentialResolver,
```

packages/vc-verification/src/issuer-verification.ts
```diff
@@ -0,0 +1,78 @@
...
+   * let issuer : 'did:ethr:volta:0x...';
+   * const role = 'role';
+   * await issuerVerification.verifyIssuer(issuer, role);
+   * ```
+   *
+   * @param issuer issuer DID to verify authority
+   * @param role authoritative role to be issuer
+   * @returns
+   */
+  async verifyIssuer(issuer: string, role: string) {
```
JGiter, line 51: Could you implement this similarly to https://github.com/energywebfoundation/ew-credentials/blob/55b84926b85a7e2261fffad8231821736d5febb6/packages/vc-verification/src/vc-issuer-verification.ts#L38? It would be more readable

packages/vc-verification/test/vc-issuer-verification.ts
```diff
@@ -341,7 +342,8 @@ function testSuite() {
 
   describe('Issuer verification', () => {
     it('verifies issuer, where the role is issued by did', async () => {
-      let ipfsCID = await didStore.save(JSON.stringify(adminVC));
+      const adminJWT = new JWT(adminKeys);
+      let ipfsCID = await didStore.save(await adminJWT.sign(adminVC));
```
JGiter, line 346: Could you clarify what is the purpose of additional signing?

JGiter (changes requested):

packages/vc-verification/src/issuer-verification.ts
```diff
@@ -0,0 +1,78 @@
...
+   * );
+   *
+   * let issuer : 'did:ethr:volta:0x...';
+   * const role = 'role';
+   * await issuerVerification.verifyIssuer(issuer, role);
+   * ```
+   *
+   * @param issuer issuer DID to verify authority
+   * @param role authoritative role to be issuer
+   * @returns
```
jrhender, line 49: Could you add description about the return value but also the expected errors that one could get when using the function? Otherwise, one needs to look into the code implementation to know the possible errors.

packages/vc-verification/src/issuer-verification.ts
nichonien: Do you mean the error handling? Or the call structure?

packages/vc-verification/src/issuer-verification.ts
```diff
@@ -0,0 +1,78 @@
...
+        return await this.vcIssuerVerification.verifyIssuer(issuer, role);
+      }
+      return await this.claimIssuerverification.verifyIssuer(issuer, role);
+    } else {
+      return issuers?.did?.find(
+        (d) => addressOf(d).toUpperCase() === addressOf(issuer).toUpperCase()
+      )
+        ? true
+        : false;
+    }
```
jrhender, line 76: I think this API is strange because if:
- Verification fails in the `did` issuer case then `false` is returned
- Verification fails in the `roleName` issuer then Errors are thrown
This should be unified. My vote would be for returning `true`/`false` with an errors array rather than throwing Errors

packages/vc-verification/test/vc-issuer-verification.ts
nichonien: This would align with the way `OffChainClaim` are published as they are signed before being published.

packages/vc-verification/src/claim-issuer-verification.ts
```diff
@@ -127,11 +127,11 @@ export class ClaimIssuerVerification {
...
-      for (let i = 0; i < issuers.did.length; i++) {
-        if (issuers.did[i] === issuerDID) {
-          return true;
-        }
-      }
+      return issuers?.did?.find(
+        (d) => addressOf(d).toUpperCase() === addressOf(issuerDID).toUpperCase()
+      )
+        ? true
+        : false;
```
jrhender, line 134: Does this also need to be changed in `vc-issuer-verification`?

I think similar to my other question, this might be creating a strange API where sometime it returns `false` on failed verification and sometimes it throws an error

jrhender (commented):
Looks like a good start but some potential tweaks needed

packages/vc-verification/src/issuer-verification.ts
JGiter: I think of them would be beneficial

packages/vc-verification/test/vc-issuer-verification.ts
JGiter: I think we should change our implementation instead of forcing user to create unnecessary signature

packages/vc-verification/src/issuer-verification.ts
JGiter: I also for consistency in return type, but I would prefer just to throw, because method returns value only to use it in some other place of code, while we are not using this value except of throwing either in library or in front-end

jrhender (commented):
Looking good to me, though I agree with @JGiter that we probably won't need to wrap the VC in a JWT when storing in IPFS

packages/vc-verification/src/Ipfs-credential-resolver.ts
```diff
@@ -142,6 +142,10 @@ export class IpfsCredentialResolver implements CredentialResolver {
       await Promise.all(
         services.map(async ({ serviceEndpoint }) => {
           const claimToken = await this._ipfsStore.get(serviceEndpoint);
+          let offChainClaim;
```
JGiter, line 145: I don't see where it is used

packages/vc-verification/src/Ipfs-credential-resolver.ts
```diff
@@ -178,8 +182,11 @@ export class IpfsCredentialResolver implements CredentialResolver {
       await Promise.all(
         services.map(async ({ serviceEndpoint }) => {
           const credential = await this._ipfsStore.get(serviceEndpoint);
-          const vc = JSON.parse(credential);
-          delete vc.iat;
+          let vc;
+          if (!(credential.split('.').length === 3)) {
```
JGiter, line 186: Does this handles non-jwt credential? How jwt credential is handled?

packages/vc-verification/src/Ipfs-credential-resolver.ts
nichonien: It stores the payload after decoding, which is followed after checking if the string fetched from IPFS is a JWT.

packages/vc-verification/src/Ipfs-credential-resolver.ts
nichonien: Yes, so to get a Verifiable Credential the string coming from IPFS should have VC. The JWT credential is handled if you fetch the OffChainClaim.

packages/vc-verification/src/Ipfs-credential-resolver.ts
JGiter: and what happens with payload next?

nichonien: My bad, I forget to return it. The old code wasn't removed, will do it.

jrhender (commented):
Just a question about `delete iat`

packages/vc-verification/src/errors.ts
```diff
@@ -73,3 +73,16 @@ export class NoIssuers extends Error {
     super(`Issuers for credential ${credential} were not set`);
   }
 }
+
+export const ERRORS = {
+  NoCredential: 'No authorative credential found for issuer',
```
JGiter, line 78: Why not get these messages from errors.ts?

packages/vc-verification/src/Ipfs-credential-resolver.ts
```diff
@@ -178,8 +182,11 @@ export class IpfsCredentialResolver implements CredentialResolver {
...
+            vc = JSON.parse(credential);
+            delete vc.iat;
```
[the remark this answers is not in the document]
nichonien: Haven't checked it recently, it is weird but in the initial testing there was `iat` property coming from IPFS though it wasn't in the sample VC. Will check it again and remove if it's not the case now.

packages/vc-verification/src/models/verifiable-credentials.ts
```diff
@@ -7,8 +7,16 @@ const { isHexString } = utils;
 
 export interface VerificationResult {
   status: boolean;
+  error: string;
 }
 
+export const verificationResult = function (
```
JGiter, line 13: I think this function isn't really needed as it just returns passed in values

JGiter (changes requested):

packages/vc-verification/src/Ipfs-credential-resolver.ts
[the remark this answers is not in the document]
jrhender: Ok, not a big deal, no need to spend much time on it. Just add a comment to explain why it is there.

packages/vc-verification/src/models/verifiable-credentials.ts
nichonien: Yeah, but if in future you want to have a different format of result Object then you just need to change it at one place.

packages/vc-verification/src/errors.ts
nichonien: Like, individual errors? I think it is easy to get all the errors throw single ERROR JSON obj and this becomes easy to export as well (no need to import individually).

packages/vc-verification/src/models/verifiable-credentials.ts
JGiter: I don't think this part is subject for changes. In any case it is not a big deal, so let's keep it now

packages/vc-verification/src/errors.ts
JGiter: I would import them since they are same errors, but it is up to you

JGiter (approved):

PR #44, opened 2022-07-13
feat(jwt-credential): replace offchainclaim interface with eip191jwt

nichonien:
This PR consists of changes to replace `OffChainClaim` with `RoleEIP191Jwt` which consists of Jwt Payload and issued Token.


jrhender (approved):
Looks good to me, thanks @nichonien

PR #46, opened 2022-07-15
feat: add revocation verification for issuer credential

nichonien:
This PR includes changes to add revocation status check for issuer's credential.

### DIFF
--- a/packages/vc-verification/src/claim-issuer-verification.ts
+++ b/packages/vc-verification/src/claim-issuer-verification.ts
@@ -1,4 +1,3 @@
-import * as jwt from 'jsonwebtoken';
 import { providers } from 'ethers';
 import { ProofVerifier } from '@ew-did-registry/claims';
 import { addressOf, Resolver } from '@ew-did-registry/did-ethr-resolver';
--- a/packages/vc-verification/src/revocation-verification.ts
+++ b/packages/vc-verification/src/revocation-verification.ts
@@ -6,17 +6,19 @@ import {
   VerifiableCredential,
 } from '@ew-did-registry/credentials-interface';
 import {
-  VCIssuerVerification,
-  ClaimIssuerVerification,
   CredentialResolver,
   IssuerResolver,
   RevokerResolver,
 } from '.';
+import { ClaimIssuerVerification } from '../src/claim-issuer-verification';
+import { VCIssuerVerification } from '../src/vc-issuer-verification';
 import { InvalidRevokerType, NoRevokers, RevokerNotAuthorized } from './errors';
 import { issuerDID, RoleEIP191JWT } from './models';
 import { addressOf } from '@ew-did-registry/did-ethr-resolver';
 import { StatusListEntryVerification } from '@ew-did-registry/revocation';
 import { RoleCredentialSubject } from '@energyweb/credential-governance';
+import { RegistrySettings } from '@ew-did-registry/did-resolver-interface';
+import { providers } from 'ethers';
 
 /**
  * Provides verification of revocation of EnergyWeb role verifiable credential
@@ -31,13 +33,22 @@ export class RevocationVerification {
     private revokerResolver: RevokerResolver,
     private issuerResolver: IssuerResolver,
     credentialResolver: CredentialResolver,
-    vcIssuerVerification: VCIssuerVerification,
-    claimIssuerVerification: ClaimIssuerVerification,
+    provider: providers.Provider,
+    registrySetting: RegistrySettings,
     private verifyProof: (vc: string, proof_options: string) => Promise<any>
   ) {
     this.credentialResolver = credentialResolver;
-    this.vcIssuerVerification = vcIssuerVerification;
-    this.claimIssuerVerification = claimIssuerVerification;
+    this.vcIssuerVerification = new VCIssuerVerification(
+      issuerResolver,
+      credentialResolver,
+      verifyProof
+    );
+    this.claimIssuerVerification = new ClaimIssuerVerification(
+      provider,
+      registrySetting,
+      credentialResolver,
+      issuerResolver
+    );
     this._statusListEntryVerificaiton = new StatusListEntryVerification(
       verifyProof
     );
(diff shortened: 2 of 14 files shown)


packages/vc-verification/src/claim-issuer-verification.ts
```diff
@@ -136,4 +159,39 @@ export class ClaimIssuerVerification {
...
     return claim instanceof Object;
   }
+
+  /**
+   * Checks the revocation status for the given StatusList2021Entry and role
+   * @param statusListEntry credentialStatus associated with the VC
+   * @param role namespace
+   * @returns
+   */
+  async checkRevocationStatus(
```
jrhender, line 169: I think there is some duplication with this `checkRevocationStatus` method and the `RevocationVerification` class. I think instead of this method, this `ClaimIssuerVerification` class should instantiate a `RevocationVerification` instance. In that way, our revocation check logic isn't duplicated. 

The same is the case in `VcIssuerVerification`

Does this make sense to you @nichonien @JGiter ?

jrhender (changes requested):

packages/vc-verification/src/claim-issuer-verification.ts
JGiter: Yes, revocation verification is more appropriate for this

nichonien: Yeah, I did that initially. But then I realised there is a circular dependency as `RevocationVerification` takes `ClaimIssuerVerification` and `VCIssuerVerification`. So instantiating `RevocationVerification` in `IssuerVerification` didn't seem the right way. Or am I missing something @jrhender @JGiter ?

jrhender: Hmm, I don't think there is a circular dependency issue 🤞 . Did you already try it out and it failed? It's going to be challenging to do it in `ClaimIssuerVerification` and `VCIssuerVerification` but in `IssuerVerification`, I think it is clearer how it is possible.

This is the current `IssuerVerification` constructor:
```
constructor(
    private issuerResolver: IssuerResolver,
    private credentialResolver: CredentialResolver,
    vcIssuerVerification: VCIssuerVerification,
    claimIssuerverification: ClaimIssuerVerification
  ) {
    this.vcIssuerVerification = vcIssuerVerification;
    this.claimIssuerverification = claimIssuerverification;
  }
```
I think one could just add another property
```
constructor(
    private issuerResolver: IssuerResolver,
    revokerResolver: RevokerResolver,
    private credentialResolver: CredentialResolver,
    vcIssuerVerification: VCIssuerVerification,
    claimIssuerverification: ClaimIssuerVerification
  ) {
    this.vcIssuerVerification = vcIssuerVerification;
    this.claimIssuerverification = claimIssuerverification;
    this.revocationVerification = new RevocationVerification(revokerResolver, credentialResolver, vcIssuerVerification, claimIssuerVerification)
  }
```

Regarding `ClaimIssuerVerification` and `VCIssuerVerification`, I think we can perhaps make `claimIssuerVerification` and `vcIssuerVerification` optional parameters to `RevocationVerification` and then `ClaimIssuerVerification` and `VCIssuerVerification` can just provide the one they have. For example,

```
constructor(
    private issuerResolver: IssuerResolver,
    private revocationResolver: RevocationResolver,
    private credentialResolver: CredentialResolver,
    private verifyProof: (vc: string, proof_options: string) => Promise<any>
  ) {
   this.revocationVerification = new RevocationVerification(revokerResolver, credentialResolver, claimIssuerVerification: this)
}
```

nichonien: @jrhender If we do that in `IssuerVerification`, `IssuerVerification` calls respective `ClaimIssuerVerification` and `VCIssuerVerification` for verification of the issuer chain based on the credential fetch for the given issuer. Now, if we want to verify the revocation status of each credential in the chain then it needs to call the revocation verification code internally and thus again the same approach. But let me try it once.

jrhender: >But let me try it once.

@nichonien Ok, thanks. let me know how it goes

packages/vc-verification/src/revocation-verification.ts
```diff
@@ -122,4 +133,61 @@ export class RevocationVerification {
       throw new InvalidRevokerType(role, revokers?.revokerType);
     }
   }
+
+  /**
+   * Checks the revocation status for the given StatusList2021Entry and role
+   * @param statusListEntry credentialStatus associated with the VC
```
jrhender, line 139:
```suggestion
   * Checks the revocation status for the given issuer and role
   * @param issuer issuer DID
```

jrhender (commented):
@nichonien I think this is looking pretty good! Some questions:

- Can you update the README with the changes from this PR please?
- It seems like we won't be checking revocation status in `ClaimIssuerVerification` and `VC IssuerVerification`. Is this right? If so, I think this is okay, but I think we shouldn't export them from the package then and should only export the `IssuerVerification` class (as the other classes could be very confusing to use if they don't check revocation). What do you think of this?
- Do we have any tests for the negative case of an issuance verification failing because the issuers credential is revoked? I didn't see any when I was looking

packages/vc-verification/test/issuer-verification-test.ts
```diff
@@ -514,10 +540,191 @@ function testSuite() {
...
+      nock(claimAdmin.credentialStatus.statusListCredential)
+        .get('')
+        .reply(200, adminStatusList);
+
+      expect(
+        (await issuerVerification.verifyIssuer(adminDid, managerRole)).status
+      ).to.be.false;
+    });
+
+    it('RoleEIP191Jwt verification should thorw, if the credential is revoked by unauthorised revoker', async () => {
```
jrhender, line 640:
```suggestion
    it('RoleEIP191Jwt verification should throw, if the credential is revoked by unauthorised revoker', async () => {
```

jrhender (approved):
Nice work, let's get it in!

packages/vc-verification/src/revocation-verification.ts
```diff
@@ -122,4 +146,61 @@ export class RevocationVerification {
...
     }
   }
+
+  /**
+   * Checks the revocation status for the given issuer and role
+   * @param issuer issuer DID
+   * @param role namespace
+   * @returns
+   */
+  async checkRevocationStatus(issuer: string, role: string) {
```
jrhender, line 156: In my opinion, it is really nice to have the return types to make it easier to read the code
```suggestion
  async checkRevocationStatus(issuer: string, role: string): Promise<Boolean> {
```

jrhender (approved):
I think we can merge as is ✅ but left a couple comments

PR #47, opened 2022-07-19
task/EC-56 different domains from resolver and registry

JGiter:
Fixes

- not include parent in subdomains from registry
- domain always read from domain reader

Also simplified getting from registry. Queueing of domains is not necessary, because queue always consists from one element

### DIFF
--- a/packages/credential-governance/src/domain-hierarchy.ts
+++ b/packages/credential-governance/src/domain-hierarchy.ts
@@ -75,21 +75,15 @@ export class DomainHierarchy {
     domain: string;
     mode: 'ALL' | 'FIRSTLEVEL';
   }): Promise<string[]> => {
-    if (!domain) throw new Error('You need to pass a domain name');
+    if (!domain) throw new Error('You need to pass a domain name'); // ?
 
-    // Filter out apps and roles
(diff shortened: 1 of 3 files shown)


packages/credential-governance/src/domain-hierarchy.ts
```diff
@@ -159,15 +150,14 @@ export class DomainHierarchy {
...
+   * using the resolver contract because each RPC call returns only direct subdomains
    */
   public getSubdomainsUsingRegistry = async ({
     domain,
   }: {
     domain: string;
   }): Promise<string[]> => {
-    if (!domain) throw new Error('You need to pass a domain name');
-    const notRelevantDomainEndings = ['roles', 'apps'];
+    if (!domain) throw new Error('You need to pass a domain name'); // ?
```
nichonien, line 160:
```suggestion
    if (!domain) throw new Error('You need to pass a domain name'); 
```
Is this intentional?

JGiter: Sorry, I forgot to remove this

nichonien (approved):

Harasz (approved):

PR #49, opened 2022-07-22
feat: validate credential expiration

nichonien:
This PR add validation for credential expiration for `RoleEIP191Jwt`.

### DIFF
--- a/packages/vc-verification/src/issuer-verification.ts
+++ b/packages/vc-verification/src/issuer-verification.ts
@@ -90,10 +90,10 @@ export class IssuerVerification {
       if (!issuerCredential) {
         return verificationResult(false, ERRORS.NoCredential);
       }
-      if (
-        !(await this.revocationVerification.checkRevocationStatus(issuer, role))
-      ) {
-        return verificationResult(false, ERRORS.IssuerCredentialRevoked);
+      const revocationStatusResult =
+        await this.revocationVerification.checkRevocationStatus(issuer, role);
+      if (!revocationStatusResult.status) {
+        return revocationStatusResult;
       }
       if (isVerifiableCredential(issuerCredential)) {
         return await this.vcIssuerVerification.verifyIssuer(issuer, role);
--- a/packages/vc-verification/test/issuer-verification-test.ts
+++ b/packages/vc-verification/test/issuer-verification-test.ts
@@ -30,9 +30,8 @@ import {
   IssuerResolver,
   RevocationVerification,
   EthersProviderRevokerResolver,
+  RevokerResolver,
 } from '../src';
-import { ClaimIssuerVerification } from '../src/claim-issuer-verification';
-import { VCIssuerVerification } from '../src/vc-issuer-verification';
 import {
   DIDAttribute,
   ProviderTypes,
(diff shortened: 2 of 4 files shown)


packages/vc-verification/test/issuer-verification-test.ts
```diff
@@ -726,5 +726,175 @@ function testSuite() {
...
+          statusListCredential:
+            'https://identitycache-dev.energyweb.org/v1/status-list/urn:uuid:4fb4e120-a566-499c-85fb-47bb5abd3d6b',
+        },
+        signer: adminDid,
+      };
+      let token: string = '';
+      let ipfsCIDAdmin: string = 'ipfsUrl';
+      if (admin.privateKey) {
+        token = await adminJWT.sign(claimAdmin);
+        if (token) {
```
JGiter, line 750: What is the case when `sign` returns nothing?

packages/vc-verification/test/issuer-verification-test.ts
```diff
@@ -726,5 +726,175 @@ function testSuite() {
...
+          statusPurpose: 'revocation',
+          statusListIndex: '0',
+          statusListCredential:
+            'https://identitycache-dev.energyweb.org/v1/status-list/urn:uuid:4fb4e120-a566-499c-85fb-47bb5abd3d6b',
+        },
+        signer: adminDid,
+      };
+      let token: string = '';
+      let ipfsCIDAdmin: string = 'ipfsUrl';
+      if (admin.privateKey) {
```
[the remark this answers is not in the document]
nichonien: Initially it use to throw error, I used the previous tests with some changes wrt to new ones. I'll confirm if it is still the same case or not and remove the check.

packages/vc-verification/test/issuer-verification-test.ts
nichonien: Will remove the check, it was because of the parent check for privateKey.

packages/vc-verification/src/issuer-verification.ts
```diff
@@ -90,10 +90,10 @@ export class IssuerVerification {
       if (!issuerCredential) {
         return verificationResult(false, ERRORS.NoCredential);
       }
-      if (
-        !(await this.revocationVerification.checkRevocationStatus(issuer, role))
-      ) {
-        return verificationResult(false, ERRORS.IssuerCredentialRevoked);
+      const revocationStatusResult =
+        await this.revocationVerification.checkRevocationStatus(issuer, role);
+      if (!revocationStatusResult.status) {
```
jrhender, line 95: I find this line pretty confusing because I typically expect `status` to be an enum of some kind instead of a boolean (e.g. HTTP status code are are large list)

This is probably a different task, but what do you think of either:
- renaming `status` to `verified` and keeping it as a boolean
- changing `status` to an enum of possible values (e.g. `VERIFIED`, `NOT_VERIFIED`)
- or some other option

jrhender (approved):
Left a comment about improving the `VerificationResult` but I think this PR looks great!

JGiter (approved):

packages/vc-verification/src/issuer-verification.ts
nichonien: Hmm, renaming it to `verified` sounds good. IMO `boolean` fits the return type and eases the validation upon it.

PR #53, opened 2022-08-04
fix(exp): update documentation to reflect expiry in milliseconds

whitneypurdum:
- Update defaultValidityPeriod documentation to make clear that time period is measured in milliseconds. (I think this is what this task was for but could be missing something!)

### DIFF
--- a/packages/credential-governance/src/types/domain-definitions.ts
+++ b/packages/credential-governance/src/types/domain-definitions.ts
@@ -25,7 +25,7 @@ export interface IRoleDefinitionV2 extends IRoleDefinitionText {
   issuer: IIssuerDefinition;
   revoker: IRevokerDefinition;
   enrolmentPreconditions: { type: PreconditionType; conditions: string[] }[];
-  // Default seconds value after which the role is considered expired, if empty then no expiry
+  // Default time period in milliseconds, after which the role is considered expired. If null or undefined, there is no default expiry
   defaultValidityPeriod?: number;
 }
 
--- a/packages/onchain-claims/test/test_utils/role-utils.ts
+++ b/packages/onchain-claims/test/test_utils/role-utils.ts
@@ -6,7 +6,7 @@ import { ClaimsRevocationRegistry as RevocationRegistry } from '../../ethers/Cla
 const { solidityKeccak256, defaultAbiCoder, arrayify } = utils;
 
 export const defaultVersion = 1;
-const expiry = Math.floor(new Date().getTime() / 1000) + 60 * 60;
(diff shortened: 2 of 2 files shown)


whitneypurdum:
@nichonien @jrhender This may be irrelevant - but it looks like expiry (Which I think is for on-chain credentials) is in seconds? Just double checking no issue with validity period and expiration date being in ms, but expiry in seconds, if this is the case? 
https://github.com/energywebfoundation/ew-credentials/blob/0d4b5036b1d1615427b25b1fc7b7629460e4f069/packages/onchain-claims/test/test_utils/role-utils.ts#L9

jrhender:
@whitneypurdum Good point about the expiry being in seconds in the line you linked to. Are we using seconds or milliseconds when we set the expiry in ICL? I think we should update this documentation to make this clear https://github.com/energywebfoundation/iam-client-lib/blob/ff6886c517bc55195691d1d316b770afa4dffde7/src/modules/claims/claims.types.ts#L201

whitneypurdum:
> @whitneypurdum Good point about the expiry being in seconds in the line you linked to. Are we using seconds or milliseconds when we set the expiry in ICL? I think we should update this documentation to make this clear https://github.com/energywebfoundation/iam-client-lib/blob/ff6886c517bc55195691d1d316b770afa4dffde7/src/modules/claims/claims.types.ts#L201

@jrhender milliseconds. I will update.

jrhender:
@whitneypurdum ok, if we are using milliseconds in ICL already then we seem to be using milliseconds pretty consistently. Maybe we should remove the `/ 1000` in the line that you linked to? Presumably that would break a test but I'm curious if that's the case.

nichonien:
> > @whitneypurdum ok, if we are using milliseconds in ICL already then we seem to be using milliseconds pretty consistently. Maybe we should remove the `/ 1000` in the line that you linked to? Presumably that would break a test but I'm curious if that's the case.
> 
> @jrhender Maybe - @nichonien can you confirm if you are using seconds for expiry? I have only worked with expirationTimestamp on the ICL side, nothing in this package, so not 100% sure what is going on or assumed.

@whitneypurdum There is no calculation of `expirationTimestamp` in `ew-credential`. The `ClaimManager` takes and store the `expiry` as it is. The pointed code is for tests and I guess we didn't really looked at it while writing. So, no change is needed here AFAIK except for the comment you have made and tests util you pointed..

whitneypurdum:
> 

Ok great, just wanted to check. Thank you!

nichonien (approved):

jrhender (approved):

PR #60, opened 2022-08-09
Sync with master

JGiter:


jrhender:
Is this required after each push to `master` or only this once for some reason?

JGiter:
This sync is exception. For some reason sync master -> develop wasn't happen on master publish

jrhender:
You mean there was some error on the previous publish that caused the sync to fail?

JGiter:
May be sync was canceled by `cancel-previous` action?

jrhender:
Ah yes, I can see that the `sync to master` action was cancelled. https://github.com/energywebfoundation/ew-credentials/runs/7725413645?check_suite_focus=true do you know why this was?

jrhender (approved):
We can give it a try. If it ends up needs to be a repeated task, we can investigate then

JGiter:
> Ah yes, I can see that the sync to master action was cancelled. https://github.com/energywebfoundation/ew-credentials/runs/7725413645?check_suite_focus=true do you know why this was?

I believe it was canceled by v2.0.0 action https://github.com/energywebfoundation/ew-credentials/actions/workflows/deploy.yml?query=branch%3Amaster

PR #62, opened 2022-08-15
task/EC-64 handle domain read errors

JGiter:


packages/credential-governance/src/errors/chain-is-not-supported.ts
```diff
@@ -0,0 +1,5 @@
+export class ChainIsNotSupported extends Error {
```
nichonien, line 1: @JGiter Can we not put all the error classes in a single file? Is there any specific reason put them separate?

JGiter: I put them separately in order to not increase file size while adding new errors

nichonien (approved):

PR #63, opened 2022-08-16
bug/SWTCH-1390 role type issuer can't issue

JGiter:


nichonien (approved):

packages/vc-verification/src/resolver/Ipfs-credential-resolver.ts
```diff
@@ -108,11 +108,13 @@ export class IpfsCredentialResolver implements CredentialResolver {
     namespace: string
   ): Promise<RoleEIP191JWT | undefined> {
     const eip191Jwts = await this.eip191JwtsOf(did);
-    return eip191Jwts.find(
-      (jwt) =>
-        jwt?.payload?.claimData.claimType === namespace ||
-        utils.namehash(jwt?.payload?.claimData.claimType) === namespace
-    );
+    return eip191Jwts
+      .filter((jwt) => typeof jwt.payload.claimData.claimType === 'string')
```
jrhender, line 112: @JGiter Thanks for suggesting the fix but I don't really get it. Could you provide an example of a jwt for which `jwt.payload.claimData.claimType === 'string'` isn't a string? I think the best way to do this would be to add a test for it.

jrhender (changes requested):
This should have a test that demonstrates the bug, IMO

packages/vc-verification/src/resolver/Ipfs-credential-resolver.ts
JGiter: This check mainly supposed to ensure that `claimType` is defined. I'll add test for this

packages/vc-verification/test/issuer-verification-test.ts
```diff
@@ -627,6 +627,78 @@ function testSuite() {
...
+        },
+      };
+      await adminOperator.update(
+        adminDid,
+        DIDAttribute.ServicePoint,
+        updateDataAdmin,
+        validity
+      );
+
+      const claimManager = {
```
jrhender, line 663: Seems like an odd name for the variable. What do you think of this instead?
```suggestion
      const managerRoleClaim = {
```

packages/vc-verification/test/issuer-verification-test.ts
```diff
@@ -627,6 +627,78 @@ function testSuite() {
       ).to.be.true;
     });
 
+    it('invalid RoleEIP191Jwt should not break verification,', async () => {
```
jrhender, line 630: I think it's better to be a bit more specific about what is invalid about the RoleEIP191Jwt. Is it that the `claimType` field is invalid?

nichonien: Yeah, but now I think this should change and If the RoleEIP191Jwt is not valid then it should break.

packages/vc-verification/src/resolver/Ipfs-credential-resolver.ts
jrhender, line 112: Would it make sense to use this helper function here? https://github.com/energywebfoundation/ew-credentials/blob/7475c9a77d5ead021ef5069a62ca315764598270/packages/vc-verification/src/utils/formatters.ts#L96

nichonien: The `isEIP191Jwt` method doesn't validate the case which Dmitry is trying to check. Maybe we make some modifications to `isEIP191Jwt` and then use it here.

jrhender: > Maybe we make some modifications to isEIP191Jwt and then use it here.

What do you think of this approach? To me this would make sense because I think this new line is just an additional check to determine if it is an `EIP191Jwt` or not

packages/vc-verification/test/issuer-verification-test.ts
```diff
@@ -627,6 +627,78 @@ function testSuite() {
       ).to.be.true;
     });
 
+    it('RoleEIP191Jwt without claimType should not break verification,', async () => {
+      const adminJWT = new JWT(adminKeys);
+      const claimAdmin = {
+        claimData: { fields: {}, claimTypeVersion: 1 },
```
jrhender, line 633: @JGiter Is this the key line in the test then? This `claimData` is missing a `claimType`?

packages/vc-verification/test/issuer-verification-test.ts
```diff
@@ -627,6 +627,78 @@ function testSuite() {
...
+      const ipfsCIDManager = await didStore.save(tokenManager);
+      const serviceIdManager = managerRole;
+      const updateDataManager: IUpdateData = {
+        type: DIDAttribute.ServicePoint,
+        value: {
+          id: `${managerDid}#service-${serviceIdManager}`,
+          type: 'ClaimStore',
+          serviceEndpoint: ipfsCIDManager,
+        },
+      };
```
jrhender, line 686: Would be good to apply the same naming convention here as well IMO. I think it is more readable. Having the adjective first is more intuitive, IMO.
```suggestion
      const managerToken = await adminJWT.sign(managerRoleClaim);
      const managerTokenCID = await didStore.save(tokenManager);
      const managerServiceId = managerRole;
      const managerUpdateData: IUpdateData = {
        type: DIDAttribute.ServicePoint,
        value: {
          id: `${managerDid}#service-${serviceIdManager}`,
          type: 'ClaimStore',
          serviceEndpoint: ipfsCIDManager,
        },
      };
```

packages/vc-verification/test/issuer-verification-test.ts
JGiter: yes, and this test breaks others

packages/vc-verification/test/issuer-verification-test.ts
jrhender: Well I think `breaking` implies that the code isn't functioning as expected, which is not what we want. I think what we want to express is that the `verified` return value should be false

JGiter: Yes, this test should ensure that verification returns true or false regardless if there are invalid among claims. verifyIssuer should not throw

packages/vc-verification/src/resolver/Ipfs-credential-resolver.ts
JGiter: I think we need dome ultimate solution. I've gave a try to ts-interface-checker/ts-interface-builder. This time I generated types manually, but we can easy automate this. What do think of this approach?

nichonien (approved):

packages/vc-verification/src/models/credentials-types-ti.ts
```diff
@@ -0,0 +1,39 @@
+/**
+ * This module was automatically generated by `ts-interface-builder`
```
jrhender, line 2: @JGiter Could you elaborate on how this automatic generate was performed? Did you need to manually trigger the generation? How will this file be updated in the future?

nichonien (dismissed):

jrhender (commented):
Some questions about how the type file generation works.

You were mentioning as well that this `ti` approach requires two files to work? For example in this case we have `credential-types-ti.ts` and `credential-types.ts`. Could you elaborate on this please? 

@JGiter as this `type-interface` approach, though maybe better long term, it might require some discussion to understand its pros/cons.
Maybe we could revert to your simpler, original fix and then discuss the interface checking approach in a separate PR. 
I'm just cognizant of trying to get the fix out to SB issues.

packages/vc-verification/src/models/credentials-types-ti.ts
JGiter: This file was generated by ts-interface-builder. I din't include generation in build script, but it is possibly to automate re-generation

JGiter: Ok, we can choose appropriate interface checker in dedicated PR. I'll revert to initial fix

packages/vc-verification/src/utils/formatters.ts
```diff
@@ -6,6 +6,7 @@ import type { RoleCredentialSubject } from '@energyweb/credential-governance';
 import type { VerifiableCredential } from '@ew-did-registry/credentials-interface';
 import { Chain } from '@ew-did-registry/did';
 import { utils } from 'ethers';
+import _ from 'lodash';
```
jrhender, line 9: Can the lodash import be optimized a bit? https://stackoverflow.com/questions/35250500/correct-way-to-import-lodash

jrhender (dismissed):
Looks good, thanks @JGiter . Let's continue the interface checker approach discussion in a separate PR 👍

packages/vc-verification/src/utils/formatters.ts
JGiter: Ah, yeah, I missed that

jrhender: Ok, no worries. I see you requested review but import is still unchanged. Is there a change incoming still here?

JGiter: Sorry. Changes were not pushed for some reason. Should be ok now

jrhender (approved):
Looks good, thanks!

PR #65, opened 2022-08-22
build: update package version

nichonien:


JGiter (changes requested):
I think we would be better merge master into dev. We will need this in any case before merge dev in master. Besides master -> dev triggers versioning from 2.1.0

nichonien:
> I think we would be better merge master into dev. We will need this in any case before merge dev in master. Besides master -> dev triggers versioning from 2.1.0

I am not sure, if this would work. IMO if you trigger this merge, it would still leave a difference in version the way it is now. Did you performed master -> dev merge before publish last time?

JGiter:
Sorry for confusion
> Besides master -> dev triggers versioning from 2.1.0

is applies to semantic-release managed repos. But we still need to sync dev with master. I did it before

nichonien:
> is applies to semantic-release managed repos. But we still need to sync dev with master. I did it before

I think this should not be the case if you already performed the master -> dev merge, the version should not be out of sync. Or would you like to give it a try again?

nichonien:
@JGiter Looking at this PR - https://github.com/energywebfoundation/ew-credentials/pull/60, I think it would be good to investigate this a bit more. To me this seems to be persistent with every new publish maybe, What do you think?

JGiter:
> To me this seems to be persistent with every new publish maybe

Yes, this needs to be fixed

JGiter (approved):

nichonien:
Closing this as https://github.com/energywebfoundation/ew-credentials/pull/69 resolved the sync issue between develop and master.

PR #74, opened 2022-09-06
feat: verify expiry in seconds

nichonien:


packages/vc-verification/src/verifier/revocation-verification.ts
```diff
@@ -205,7 +205,7 @@ export class RevocationVerification {
             issuer = rolePayload?.iss as string;
             role = issuers.roleName;
             credentialStatus = rolePayload?.credentialStatus;
-            if (rolePayload?.exp && rolePayload?.exp < Date.now()) {
+            if (rolePayload?.exp && rolePayload?.exp * 1000 < Date.now()) {
```
JGiter, line 208: Optional chaining can be skipped here
```suggestion
            if (rolePayload.exp && rolePayload.exp * 1000 < Date.now()) {
```

JGiter (approved):

packages/vc-verification/src/verifier/revocation-verification.ts
nichonien: It throws error `object possibly undefined`

PR #77, opened 2022-09-19
feat:  cache data from resolvers

nichonien:
This PR introduces caching through Data Aggregator classes from resolvers to improve efficiency for issuer verification.


packages/vc-verification/src/resolver/data-aggregator.ts
```diff
@@ -0,0 +1,54 @@
+import {
+  CachedDIDDocument,
+  CachedRoleCredential,
+  CachedRoleDefinition,
+} from '../models';
+
+/**
+ * An Interface for Data aggreator to cache responses from Blockchain and SSI-Hub
+ */
+export interface IDataAggregator {
```
JGiter: Is it supposed to be different implementations of IDataAggregator?

JGiter (commented):
Thanks for PR. There are two points not clear to me:
- how we can ensure that cached data is in sync with actual data?
- isn't it dangerous that cache accumulates all data? I thought that caches are used to speed access to most frequently requested data?

Also what do you think of caching role definition in DomainReader? It would allows to benefit from caching in all places where role is read, not just in verification

packages/vc-verification/src/resolver/data-aggregator.ts
nichonien: Yeah, there would be different implementation of IDataAggregator. So for the case in `ew-credentials` it is just the data coming from blockchain. While in `passport-did-auth` the data would be coming from both blockchain and SSI-Hub.

nichonien:
> * how we can ensure that cached data is in sync with actual data?
> * isn't it dangerous that cache accumulates all data? I thought that caches are used to speed access to most frequently requested data?

Yeah, this is a good point. So the caching should be specific to the request itself i.e. it should be cleared once any request (verification) is completed.

> Also what do you think of caching role definition in DomainReader? It would allows to benefit from caching in all places where role is read, not just in verification.

Hmm, that's a great point. I will look into this.

packages/vc-verification/src/resolver/data-aggregator.ts
JGiter: What is expected difference would be between their implementations. Is `passport-did-auth` version expected to do something else except getting and setting? 
Also name `EVMDataAggregator` is quite vague from my perspective

JGiter:
> So the caching should be specific to the request itself i.e. it should be cleared once any request (verification) is completed.

This is reduces gain from caching. It seems to me that effective caching is quite challenging task. May instead of caching here we would better do this in ssi-hub https://docs.nestjs.com/techniques/caching?

nichonien:
> This is reduces gain from caching. It seems to me that effective caching is quite challenging task. May instead of caching here we would better do this in ssi-hub https://docs.nestjs.com/techniques/caching?

I still feel this would be really efficient for issuer verification where at least four calls (either credential or role definition) for a verification of a credential issued by Role type. Also, this does keep the cache short, since a user verification is not that often (based on switchboard use case). A single user has very few interaction and it makes sense to keep the cache short.

With reg. to SSI-Hub caching, it could be helpful if the system is connected to it and fetches some data. For issuer verification it is just the blockchain for now.

packages/vc-verification/src/resolver/data-aggregator.ts
nichonien: The data type that is being stored would create difference. Since data returned from SSI-Hub are different to what comes from blockchain. 

>Also name EVMDataAggregator is quite vague from my perspective

Yeah, I didn't find any other relevant naming for this.

nichonien:
@JGiter I tried adding caching service to DomainReader:
1. We would need a sync service that listens to ENS resolvers, this would ensure the cached definitions are synced if there is an update made to domain.
2. Majorly all the definition resolution is done through SSI-Hub for switchboard, so this would not make the resolution efficient given the complexity it would because of sync service.

C.C @jrhender

packages/vc-verification/src/resolver/data-aggregator.ts
```diff
@@ -0,0 +1,50 @@
...
+  /**
+   * Stores role credential
+   * @param did user DID
+   * @param role namespace
+   * @param data Credential of the user for the namespace
+   */
+  setRoleCredential(did: string, role: string, data: any): void;
```
JGiter: Can we set exact type of data?
```suggestion
  setRoleCredential(did: string, role: string, data: RoleEIP191JWT | VerifiableCredential<RoleCredentialSubject>): void;
```

packages/vc-verification/src/resolver/evm-data-aggregator.ts
```diff
@@ -0,0 +1,91 @@
...
+import { IServiceEndpoint } from '@ew-did-registry/did-resolver-interface';
+import { IRoleDefinitionV2 } from '@energyweb/credential-governance';
+import { VerifiableCredential } from '@ew-did-registry/credentials-interface';
+import type { RoleCredentialSubject } from '@energyweb/credential-governance';
+
+/**
+ * A class to aggreagate data from EVMs
+ */
+export class EVMDataAggregator implements IDataAggregator {
+  private cahedRoleDefinition: CachedRoleDefinition = {};
```
jrhender: Small typo
```suggestion
  private cachedRoleDefinition: CachedRoleDefinition = {};
```

packages/vc-verification/src/resolver/evm-data-aggregator.ts
```diff
@@ -0,0 +1,91 @@
...
+  private cachedDIDDocument: CachedDIDDocument = {};
+  private cachedRoleCredential: CachedRoleCredential[] = [];
```
jrhender: Why an array for `cacheRoleCredential`? Seems to me like it would be more intuitive as either
- a two layer nested map, where the first key gets index by DID and second key indexes by role namespace
- at least use a map to index by did and iterate to find the matching role

packages/vc-verification/src/resolver/issuer-resolver.ts
```diff
@@ -21,9 +22,11 @@ export interface IssuerResolver {
  */
 export class EthersProviderIssuerResolver implements IssuerResolver {
   private _domainReader: DomainReader;
+  private _dataAgregator: EVMDataAggregator;
 
-  constructor(domainReader: DomainReader) {
+  constructor(domainReader: DomainReader, dataAggregator: EVMDataAggregator) {
```
jrhender: Why is `EVMDataAggregator` being passed via the constructor? Is this because it could be shared between classes?

packages/vc-verification/src/resolver/evm-data-aggregator.ts
```diff
@@ -0,0 +1,91 @@
...
+import { IDataAggregator } from './data-aggregator';
+import { IServiceEndpoint } from '@ew-did-registry/did-resolver-interface';
+import { IRoleDefinitionV2 } from '@energyweb/credential-governance';
+import { VerifiableCredential } from '@ew-did-registry/credentials-interface';
+import type { RoleCredentialSubject } from '@energyweb/credential-governance';
+
+/**
+ * A class to aggreagate data from EVMs
+ */
+export class EVMDataAggregator implements IDataAggregator {
```
jrhender: I don't think this really needs to be implementing an interface to be honest. The sign of that is that `EthersProviderIssuerResolver` is accepting a concrete `EVMDataAggregator` and not the interface.
```suggestion
export class EVMDataAggregator {
```

packages/vc-verification/src/resolver/data-aggregator.ts
jrhender: I agree with @JGiter 's point that `EVMDataAggregator` is quite vague. To me, this is because it is violating the single-responsability-principle and so it is hard to find a specific name for it. I recommend splitting the cache up into a cache for each type of cached data.

jrhender:
@nichonien 

> So the caching should be specific to the request itself i.e. it should be cleared once any request (verification) is completed.

I think the above idea is good. I also see it as been valuable to reduce server requests and I think a single "verification request" is probably the right retention policy for the cache.

One thing I'm confused by though is why it is being implemented in ew-credentials. Wouldn't it have been better to have gone straight to an implementation in passport-did-auth? Why is this implementation needed?

packages/vc-verification/src/resolver/issuer-resolver.ts
nichonien: Yes, so the same cached `roleDefinition` could be used to return both `issuer` and `revoker`.

packages/vc-verification/src/resolver/evm-data-aggregator.ts
nichonien: This is needed because the implementations in passport-did-auth would also need the same interface to pass their own implementation of Cache.

nichonien:
> One thing I'm confused by though is why it is being implemented in ew-credentials. Wouldn't it have been better to have gone straight to an implementation in passport-did-auth? Why is this implementation needed?

This is needed because the implementations in passport-did-auth would also need the same interface to pass their own implementation of Cache. Since, the resolver interfaces are in ew-credentials. Also, the resolvers in passport-did-auth use resolvers from ew-credentials for fetching data from blockchain.

packages/vc-verification/src/resolver/evm-data-aggregator.ts
jrhender: Ok, I agree with @JGiter that I don't see why passport-did-auth will need a different cache implementation, but it doesn't really hurt to have an interface

packages/vc-verification/src/resolver/issuer-resolver.ts
jrhender: The constructor should accept the interface of the cache, not the implementation.

jrhender:
@nichonien looking ahead a bit, one thing I'm a bit fuzzy on is how the cache will be on a "per-request" basis. Is it clear to you how this will work?

I'm wondering if perhaps an additional cache key to have a "per-request" cache is needed?

packages/vc-verification/src/verifier/issuer-verification.ts
```diff
@@ -77,7 +78,9 @@ export class IssuerVerification {
       | VerifiableCredential<RoleCredentialSubject>
       | RoleEIP191JWT
       | undefined;
-
+    const dataCache = new IntermediateDataCache();
+    this.credentialResolver.setCache(dataCache);
+    this.issuerResolver.setCache(dataCache);
```
JGiter: I think it is also possible to set cache in revocation-verification

packages/vc-verification/src/resolver/data-aggregator.ts
JGiter: The `IntermediaryDataCache` doens't point out what is specific in this implementation of `IDataCache`. You argued that interface introduced because data from ssi-hub differs from data from chain. Could you please specify which exact data do you mean. I thought that ssi-hub implements interfaces defined in ew-credentials

packages/vc-verification/src/resolver/intermediate-data-cache.ts
```diff
@@ -0,0 +1,89 @@
...
+import { IDataCache } from './idata-cache';
+import { IServiceEndpoint } from '@ew-did-registry/did-resolver-interface';
+import { IRoleDefinitionV2 } from '@energyweb/credential-governance';
+import { VerifiableCredential } from '@ew-did-registry/credentials-interface';
+import type { RoleCredentialSubject } from '@energyweb/credential-governance';
+
+/**
+ * A class to aggreagate data from EVMs
+ */
+export class IntermediateDataCache implements IDataCache {
```
jrhender: IMO, the `IDataCache` interface is too large and introduces Interface Segregation Principle violations. I recommend splitting it into several smaller interfaces 
```suggestion
export class IntermediateDataCache implements IRoleCache, IDIDDocCache, IRoleCredentialCache, IClearableCache {
```

packages/vc-verification/src/resolver/issuer-resolver.ts
```diff
@@ -40,15 +48,28 @@ export class EthersProviderIssuerResolver implements IssuerResolver {
...
       return roleDefinition.issuer;
     }
     return undefined;
   }
+
+  /**
+   * Sets intermediate cache
+   * @param dataCache
+   */
+  setCache(dataCache: IDataCache): void {
```
jrhender: I think this is an Interface Segregation Principle violation because `EthersProviderIssuerResolver` only needs the `RoleDefinition` cache but it is exposed to the `DIDDocument` and `RoleCredential` cache methods.
I think a way to solve this would be as follows (given interface segregation proposed in my previous comment):
```suggestion
  setRoleDefinitionCache(roleDefinitionCache: IRoleDefinitionCache): void {
```

packages/vc-verification/src/resolver/revoker-resolver.ts
```diff
@@ -40,15 +48,28 @@ export class EthersProviderRevokerResolver implements RevokerResolver {
...
     return undefined;
   }
+
+  /**
+   * Sets intermediate cache
+   * @param dataCache
+   */
+  setCache(dataCache: IDataCache): void {
+    this._dataCache = dataCache;
+  }
```
jrhender: Same ISP violation as noted in my comment in `EthersProviderIssuerResolver`

packages/vc-verification/src/resolver/idata-cache.ts
```diff
@@ -0,0 +1,50 @@
...
+   * @param data Credential of the user for the namespace
+   */
+  setRoleCredential(did: string, role: string, data: any): void;
+
+  /**
+   * Stores DID Document
+   * @param did user DID
+   * @param data DIDDocument of the user (DID)
+   */
+  setDIDDocument(did: string, data: any): void;
```
jrhender: Similar to @JGiter 's comment about `setRoleCredential`, is it possible to use an interface for a `DIDDocument` here rather than `any`?

At a minimum, if there is no suitable interface, it is better to use `unknown` than `any`, I believe: https://stackoverflow.com/questions/51439843/unknown-vs-any

packages/vc-verification/src/resolver/intermediate-data-cache.ts
jrhender: I would also consider having separate caches per object type. IMO, this would better follow the SRP and it seems the clients of the cache only needs a single cache type. E.g. `EthersProviderIssuerResolver` only needs role definitions.
So could do:
```
RoleDefinitionCache implements IRoleDefinitionCache;
DIDDocCache implemented IDIDDocCache;
RoleCredentialCache implements IRoleCredentialCache;
```

jrhender (changes requested):
IMO, there is no need to have the Interface Segregation Principle violations and it would be easy to change, so this should be done in this PR.

packages/vc-verification/src/resolver/idata-cache.ts
```diff
@@ -0,0 +1,50 @@
...
+  /**
+   * Returns cached role defintion
+   * @param role namespace
+   */
+  getRoleDefinition(role: string): any | undefined;
+
+  /**
+   * Clears all cached data
+   */
+  clearData(): void;
```
jrhender: Also, I see this `clearData` function isn't being used anywhere. IMO, from a YAGNI principle perspective, it should be removed.

packages/vc-verification/src/models/cache-interfaces.ts
```diff
@@ -0,0 +1,70 @@
+import {
+  IRoleDefinitionV2,
+  RoleCredentialSubject,
+} from '@energyweb/credential-governance';
+import { VerifiableCredential } from '@ew-did-registry/credentials-interface';
+import { RoleEIP191JWT } from './credentials-types';
+import { IDIDDocument } from '@ew-did-registry/did-resolver-interface';
+
+/**
+ * An Interface to cache DID Document from Blockchain and SSI-Hub
```
jrhender: IMO, I would keep this comment generic as the interface isn't really tied to "Blockchain" or "SSI-HUB" in any way
```suggestion
 * An Interface to cache DID Documents
```

packages/vc-verification/src/models/cache-interfaces.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  /**
+   * Returns cached DID Document (services)
+   * @param did user DID
+   */
+  getDIDDocument(did: string): IDIDDocument | undefined;
+}
+
+/**
+ * An Interface to cache role credential from Blockchain and SSI-Hub
```
jrhender: Same rationale as previous comment
```suggestion
 * An Interface to cache role credential
```

packages/vc-verification/src/models/cache-interfaces.ts
```diff
@@ -0,0 +1,70 @@
...
+   * @param role namespace
+   */
+  getRoleCredential(
+    did: string,
+    role: string
+  ): RoleEIP191JWT | VerifiableCredential<RoleCredentialSubject> | undefined;
+}
+
+/**
+ * An Interface to cache role definition from Blockchain and SSI-Hub
```
jrhender: Same rationale as previous comment
```suggestion
 * An Interface to cache role definition
```

packages/vc-verification/src/resolver/Ipfs-credential-resolver.ts
```diff
@@ -174,4 +186,12 @@ export class IpfsCredentialResolver implements CredentialResolver {
       )
     ).filter(isVerifiableCredential);
   }
+
+  /**
+   * Sets intermediate cache
```
jrhender:
```suggestion
   * Sets role credential cache
```

packages/vc-verification/src/resolver/Ipfs-credential-resolver.ts
```diff
@@ -57,10 +59,20 @@ export class IpfsCredentialResolver implements CredentialResolver {
...
+    if (cachedRoleCredential) {
+      return cachedRoleCredential;
+    }
     credential = await this.getVerifiableCredential(did, namespace);
     if (!credential) {
       credential = await this.getEIP191JWT(did, namespace);
     }
+    if (credential) {
+      this._roleCredentialCache?.setRoleCredential(did, namespace, credential);
+    }
```
jrhender: Thinking about this some more, I think this cache might be ineffective. 
It seems to be the case for each `did`+`namespace` combination, `this.credentialsOf(did)` will be called.
Maybe it makes sense to cache all of the credentials resolved for a given `did`? Though this would increase the cache size, which we might not want.
What are your thoughts here @nichonien ?

packages/vc-verification/src/models/credentials-types.ts
```diff
@@ -29,3 +33,15 @@ export interface RoleEIP191JWT {
...
+  [key: string]: RoleEIP191JWT | VerifiableCredential<RoleCredentialSubject>;
+}
+
+export interface CachedRoleDefinition {
+  [key: string]: IRoleDefinitionV2;
+}
+
+export interface CachedDIDDocument {
+  [key: string]: IDIDDocument;
+}
```
jrhender: IMO, as these interfaces are only used as the type of the private cache implementations (as far as I can tell), I think this code should be removed and the type can be set directly in the cache implementations. My rationale for this is that "the best code is no code at all".
```suggestion
```

jrhender (commented):
It's looking good! I just added a few additional suggestions.

I'm also curious why no additional tests for this code @nichonien ? Maybe it makes sense in this case, but I'm curious what your rationale is. I can see that plausibly having tests around the cache implementations (though they are very simple) could make sense or maybe a test like "can resolve the same role credential twice" could be useful around the credential-resolver class, for example.

packages/vc-verification/src/models/cache-interfaces.ts
```diff
@@ -0,0 +1,70 @@
...
+ * An Interface to cache role credential from Blockchain and SSI-Hub
+ */
+export interface IRoleCredentialCache {
+  /**
+   * Stores role credential
+   * @param did user DID
+   * @param role namespace
+   * @param data Credential of the user for the namespace
+   */
+  setRoleCredential(
```
JGiter, line 37: I would also chose more descriptive names, so that when you see method you understand on which object iit is called
```suggestion
  cacheRoleCredential(
```

packages/vc-verification/src/resolver/Ipfs-credential-resolver.ts
nichonien: @jrhender If you look at this method signature it does have both `did` and `namespace`, so it can very well pull credentials from cache. While the `this.credentialsOf(did)` is only called to fetch all the VCs and then it is processed to return VC specific to the namespace. Also, the first call will always be for `getCredential` and it is much convenient to pull cache data here itself. Am I missing something or not understanding you concern?

jrhender: My concern, in summary, is that there would be multiple calls to `credentialsOf` for the same DID to retrieve the same information if there are multiple calls to `getCredentials` for different DID+namespace combinations. Do you see this as a problem?

nichonien:
> I'm also curious why no additional tests for this code @nichonien ? Maybe it makes sense in this case, but I'm curious what your rationale is. I can see that plausibly having tests around the cache implementations (though they are very simple) could make sense or maybe a test like "can resolve the same role credential twice" could be useful around the credential-resolver class, for example.

This cache is internal to resolver, so the end result is still the same (either coming form blockchain or cache). So it is hard to verify the same through tests. Maybe I am not seeing how it can be done.

packages/vc-verification/src/resolver/Ipfs-credential-resolver.ts
nichonien: Nope, the credential resolution will never go to `credentialsOf` if we have already cached the role credential and thus `getCredential` will return without moving forward. The issuer-verification code relies only on `getCredential` method.

jrhender:
> This cache is internal to resolver,

I don't think this is the case. For instance, `RoleCredentialCache` is a separate class from `CredentialResolver`, so in principle it could be unit tested separately.
Maybe it doesn't make sense to unit test `RoleCredentialCache` that as the class is to too simple, but maybe it does. What are your thoughts?

I guess you're saying that because `RoleCredentialCache` isn't exported from the package, it doesn't need to be unit tested?

packages/vc-verification/src/resolver/Ipfs-credential-resolver.ts
```diff
@@ -61,18 +59,20 @@ export class IpfsCredentialResolver implements CredentialResolver {
...
       namespace
     );
     if (cachedRoleCredential) {
-      return cachedRoleCredential.data;
+      return cachedRoleCredential;
     }
     credential = await this.getVerifiableCredential(did, namespace);
     if (!credential) {
       credential = await this.getEIP191JWT(did, namespace);
     }
```
JGiter: I guess we can also cache non-existent credentials. For this `getRoleCredential` can return `undefined` when request was not made yet and `null` if credential was not issued

packages/vc-verification/src/resolver/Ipfs-credential-resolver.ts
jrhender: > Nope, the credential resolution will never go to credentialsOf if we have already cached the role credential

But if you request a different credential for the same DID, then it will go to `credentialsOf` again, right? That is my concern.

nichonien:
> I don't think this is the case. For instance, `RoleCredentialCache` is a separate class from `CredentialResolver`, so in principle it could be unit tested separately. Maybe it doesn't make sense to unit test `RoleCredentialCache` that as the class is to too simple, but maybe it does. What are your thoughts?

Yes, for a unit test this is too simple. So I was only thinking it to be tested with resolvers, but now I will add some unit tests.

packages/vc-verification/src/models/cache-interfaces.ts
jrhender: IMO, `set` is actually clearer than `cache`, but I would leave it up to you @nichonien

packages/vc-verification/src/resolver/Ipfs-credential-resolver.ts
nichonien: Hmm, then I think a better way to do it would be to cache the DID Document and process it for newer credentials. What do you think?

packages/vc-verification/src/resolver/Ipfs-credential-resolver.ts
nichonien: This is a nice idea, but if you consider an issuer-verification call then if the credential is not there it would return with a unverified status or throw error. So caching a non-existent credential doesn't really add a value IMO.

packages/vc-verification/src/resolver/Ipfs-credential-resolver.ts
jrhender: I think it depends on how we anticipate the cache being used. The cache for `IpfsCredentialResolver` is currently optimized for resolving the same role credential for the same DID multiple times. Is this the use case that `IpfsCredentialResolver` is used for for issuer and revoker verification?

packages/vc-verification/src/resolver/Ipfs-credential-resolver.ts
JGiter: We can throw only when there is `null`. When credential is `undefined` then we need to fetch from chain

nichonien: That behaviour is already there, if it is undefined then credential is being fetched from chain.

packages/vc-verification/src/resolver/Ipfs-credential-resolver.ts
```diff
@@ -81,6 +90,13 @@ export class IpfsCredentialResolver implements CredentialResolver {
    */
   async getVerifiableCredential(did: string, namespace: string) {
     const credentials = await this.credentialsOf(did);
+    credentials.map((credential) =>
```
jrhender: I think `forEach` is a bit more appropriate here we are not interested in the resulting array from the `map`
```suggestion
    credentials.forEach((credential) =>
```

packages/vc-verification/src/models/credentials-types.ts
jrhender: I still think these interfaces are not needed, but it's a minor point
```suggestion
```

jrhender (dismissed):
Looks good, thanks @nichonien !

jrhender (approved):

PR #81, opened 2022-09-30
fix: add stack trace to error message

whitneypurdum:
- add stack trace to revoker error message to see where .ipfs of undefined is coming from

### DIFF
--- a/packages/vc-verification/src/utils/errors.ts
+++ b/packages/vc-verification/src/utils/errors.ts
@@ -25,10 +25,18 @@ export class CredentialRevoked extends Error {
 }
 
 export class RevokerNotAuthorized extends Error {
-  constructor(revoker: string, credential: string, reason?: string) {
+  constructor(
+    revoker: string,
+    credential: string,
+    reason?: string,
+    stack?: string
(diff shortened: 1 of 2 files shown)


jrhender (approved):
Ok, let's give it a try

packages/vc-verification/src/utils/errors.ts
```diff
@@ -25,11 +25,19 @@ export class CredentialRevoked extends Error {
 }
 
 export class RevokerNotAuthorized extends Error {
-  constructor(revoker: string, credential: string, reason?: string) {
+  constructor(
+    revoker: string,
+    credential: string,
+    reason?: string,
+    stack?: string
```
nichonien, line 32: @whitneypurdum Just curious, are we expecting some unexpected error?

whitneypurdum: @nichonien this is to add a stack trace to see where ipfs is being read, or where the error is coming from. right ow we are just assuming where it is happening (in ew-did-registry). This will now provide the stack trace for the error so we can see where it is happening for sure.

PR #82, opened 2022-10-03
feat: refactor caching for resolvers

nichonien:


packages/vc-verification/src/resolver/Ipfs-credential-resolver.ts
```diff
@@ -42,33 +41,39 @@ export class IpfsCredentialResolver implements CredentialResolver {
...
    * @param did subject DID for which the credential needs to be fetched
    * @param namespace role for which the credential needs to be fetched
+   * @param roleCredentialCache Cache to store role credentials
    * @returns
    */
   async getCredential(
     did: string,
-    namespace: string
+    namespace: string,
+    roleCredentialCache: IRoleCredentialCache
```
jrhender: I think it should be optional param. In this way, someone doesn't have to provide a cache if they don't want to use one. There is already the check for it in the method implementation. If you think it is easier to not have it as an optional param though, let me know
```suggestion
    roleCredentialCache?: IRoleCredentialCache | undefined
```

packages/vc-verification/src/resolver/Ipfs-credential-resolver.ts
```diff
@@ -81,17 +86,22 @@ export class IpfsCredentialResolver implements CredentialResolver {
...
    * @param did subject DID for which the credential needs to be fetched
    * @param namespace role for which the credential needs to be fetched
+   * @param roleCredentialCache Cache to store role credentials
    * @returns
    */
-  async getVerifiableCredential(did: string, namespace: string) {
+  async getVerifiableCredential(
+    did: string,
+    namespace: string,
+    roleCredentialCache: IRoleCredentialCache
```
jrhender:
```suggestion
    roleCredentialCache?: IRoleCredentialCache | undefined
```

packages/vc-verification/src/resolver/Ipfs-credential-resolver.ts
```diff
@@ -81,17 +86,22 @@ export class IpfsCredentialResolver implements CredentialResolver {
...
+  ) {
     const credentials = await this.credentialsOf(did);
     credentials.forEach((credential) =>
-      this._roleCredentialCache?.setRoleCredential(
+      roleCredentialCache.setRoleCredential(
```
jrhender: if roleCredentialCache is optional param, an additional check for it is necessary here

packages/vc-verification/src/resolver/Ipfs-credential-resolver.ts
```diff
@@ -117,17 +127,19 @@ export class IpfsCredentialResolver implements CredentialResolver {
...
    * @param did subject DID for which the credential to be fetched
    * @param namespace role for which the credential need to be fetched
+   * @param roleCredentialCache Cache to store role credentials
    * @returns RoleEIP191JWT
    */
   async getEIP191JWT(
     did: string,
-    namespace: string
+    namespace: string,
+    roleCredentialCache: IRoleCredentialCache
```
jrhender:
```suggestion
    roleCredentialCache?: IRoleCredentialCache | undefined
```

packages/vc-verification/src/resolver/Ipfs-credential-resolver.ts
jrhender: One thing to may consider here is is whether or not the addition of the `credentials` to the cache should not done in this function and instead was left to be done after this function returns.
This is because ideally, one should be able to trust that read function (which is implied by `get`) has no side-effects.
I think the side-effect to be avoided in this case is "[modifying a mutable argument passed by reference](https://en.wikipedia.org/wiki/Side_effect_(computer_science))"
What I'm thinking is that both the one credential that is requested plus the whole set of resolved credentials can be returned as a tuple. Then, the caller can add the set of credentials to the cache if they would like.

jrhender (commented):
I think the caches can be optional params + another point about avoiding side-effects in read/get methods

packages/vc-verification/src/resolver/Ipfs-credential-resolver.ts
nichonien: So what you are referring is that we should get all credentials in the `getCredential` and then cache it? @jrhender

packages/vc-verification/src/resolver/Ipfs-credential-resolver.ts
nichonien: I also had the same though but wasn't sure if it would be a good idea to make it optional. On the other hand I think it should be made mandatory since all the exported methods do perform `issuerVerification` some way or the other, so this would improve the efficiency for sure, but it could be optional and left it upto to the caller.

jrhender: Yes, on balance, I'm leaning towards it being optional

packages/vc-verification/src/resolver/Ipfs-credential-resolver.ts
jrhender: @nichonien yes, what I'm envisioning is that `getVerifiableCredential` would be called like this
```
const cache = new CredentialCache();
const { namespaceCredential, allRetrievedCredentials } = getVerifiableCredential(did, namespace, cache);
cache.add(allRetrievedCredentials);
```
In this way, we remove the side-effect of cache modification from `getVerifiableCredential` and control/modification of the cache is fully in control of the caller. IMO, this approach is easier to reason about because there is no state change hidden in a method.

packages/vc-verification/src/verifier/claim-issuer-verification.ts
```diff
@@ -43,30 +49,43 @@ export class ClaimIssuerVerification {
...
+      ) {
         return verificationResult(false, ERRORS.IssuerNotAuthorized);
       }
       const roleIssuers = await this._issuerDefResolver.getIssuerDefinition(
-        role
+        role,
+        roleDefCache
       );
       if (roleIssuers && roleIssuers.did) {
         return verificationResult(true, '');
```
JGiter, line 83: Could you please remind why we don't check that issuer is in roleIssuers.did?

packages/vc-verification/src/verifier/revocation-verification.ts
```diff
@@ -80,16 +77,23 @@ export class RevocationVerification {
    * );
    * let credential : StatusList2021Credential;
    * const role = 'role';
-   * await revocationVerification.verifyStatusList(credential, role);
+   * await revocationVerification.verifyStatusList(credential, role, roleCredentialCache, roleDefCache);
```
JGiter:
```suggestion
   * await revocationVerification.verifyStatusList(statusList, role, roleCredentialCache, roleDefCache);
```

JGiter (commented):
I wonder if it is possible to separate caching logic from verification. What if caching was incapsulated in classes. For example
```
class CredentialResolverCached extends IpfsCredentialResolver {
   cache: ...

   getVerifiableCredential(did, namespace) {
      const cached = cache[did][namespace];
      if (cached) {
         return cached;
      else { ...}
   }
...
}
This is also allows to add features to make caching more flexible

File: packages/vc-verification/src/verifier/claim-issuer-verification.ts (hunk 5 above)
Comment (nichonien) [reply packages/vc-verification/src/verifier/claim-issuer-verification.ts:83]: So, the verification of issuer is already being done before this call. https://github.com/energywebfoundation/ew-credentials/blob/1747ee878ef05e3534dcc67717dcdd0b81515ad5/packages/vc-verification/src/verifier/claim-issuer-verification.ts#L69.

This check is to validate if the chain has ended or it it is a role based issuance then set the new issuers and role to be verified.

**Comment by nichonien** [comment]
> I wonder if it is possible to separate caching logic from verification. What if caching was incapsulated in classes. For example
> 
> ```
> class CredentialResolverCached extends IpfsCredentialResolver {
>    cache: ...
> 
>    getVerifiableCredential(did, namespace) {
>       const cached = cache[did][namespace];
>       if (cached) {
>          return cached;
>       else { ...}
>    }
> ...
> }
> This is also allows to add features to make caching more flexible
>          
> ```

To me this seems to be over do, caching and resolution seems to go hand in hand for the best outcome and the need to caching is because of the verification.

File: packages/vc-verification/src/resolver/Ipfs-credential-resolver.ts (hunk 3 above)
Comment (nichonien) [reply packages/vc-verification/src/resolver/Ipfs-credential-resolver.ts]: Working at the code more, I think it is best to cache in `getVerifiableCredential` and `getEIP191JWT`. This is because the `getCredential` method is used only twice, while the specific methods are used for entire issuer's chain interaction. So `ClaimIssuerVerification` relies on `getEIP191JWT` for all credential resolution. What do you think @jrhender

Comment (jrhender) [reply packages/vc-verification/src/resolver/Ipfs-credential-resolver.ts]: Ok, sounds good 👍

Comment (nichonien) [reply packages/vc-verification/src/resolver/Ipfs-credential-resolver.ts]: Or maybe we do it in both the methods, so the `getCredential` relies on `credentialsOf` and `eip191JwtsOf` directly. So the credentials get cached within the main call method.

**Comment by JGiter** [comment]
> To me this seems to be over do, caching and resolution seems to go hand in hand for the best outcome and the need to caching is because of the verification.

Ok. I just thought that it would allow us to leave intact code of verifies, which is already quit complicated

File: packages/vc-verification/src/resolver/Ipfs-credential-resolver.ts (hunk 3 above)
Comment (jrhender) [reply packages/vc-verification/src/resolver/Ipfs-credential-resolver.ts]: You're suggestioning to add to the cache in `credentialsOf` and `eip191JwtsOf`?

Comment (nichonien) [reply packages/vc-verification/src/resolver/Ipfs-credential-resolver.ts]: No, I was thinking to call `credentialsOf` and `eip191JwtsOf` from `getCredential` and then cache all the credentials. This way `getCredential`, `getVerifiableCredential` and `getEIP191JWT` would cache credentials whenever called, but this was increasing complexity.

File: packages/vc-verification/src/resolver/Ipfs-credential-resolver.ts [hunk 7]
```diff
@@ -81,17 +86,29 @@ export class IpfsCredentialResolver implements CredentialResolver {
    *  provider,
    *  registrySettings,
    *  didStore );
-   * const credential = credentialResolver.getVerifiableCredential('did:ethr:1234', 'sampleRole');
+   * const credential = credentialResolver.getVerifiableCredential('did:ethr:1234', 'sampleRole', roleCredentialCache);
    * ```
    *
    * @param did subject DID for which the credential needs to be fetched
    * @param namespace role for which the credential needs to be fetched
+   * @param roleCredentialCache Cache to store role credentials
```
jrhender: I think it would be good to mention that the cache is populated (to at least communicate the side-effect)
```suggestion
   * @param roleCredentialCache Cache to store role credentials. Cache is updated with all credentials retrieved for the DID
```

packages/vc-verification/src/resolver/Ipfs-credential-resolver.ts
```diff
@@ -117,17 +134,26 @@ export class IpfsCredentialResolver implements CredentialResolver {
    *
    * @param did subject DID for which the credential to be fetched
    * @param namespace role for which the credential need to be fetched
+   * @param roleCredentialCache Cache to store role credentials
```
jrhender:
```suggestion
   * @param roleCredentialCache Cache to store role credentials. Cache is updated with all credentials retrieved for the DID
```

packages/vc-verification/src/resolver/issuer-resolver.ts
```diff
@@ -10,25 +10,20 @@ export interface IssuerResolver {
   /**
    * Fetches authorised issuers for the provided namespace
    * @param namespace for which the issuers needs to be fetched
+   * @param roleDefCache Cache to store and fetch RoleDefinition
```
jrhender:
```suggestion
   * @param roleDefCache Cache to store and fetch RoleDefinition. Cache is updated with retrieved role definition if not present.
```

packages/vc-verification/src/resolver/issuer-resolver.ts
```diff
@@ -40,16 +35,17 @@ export class EthersProviderIssuerResolver implements IssuerResolver {
    * ```typescript
    * const issuerResolver = new EthersProviderIssuerResolver(domainReader);
    * const role = 'sampleRole';
-   * const issuers = issuerResolver.getIssuerDefinition(sampleRole);
+   * const issuers = issuerResolver.getIssuerDefinition(sampleRole, roleDefCache);
    * ```
    * @param namespace for which the issuers need to be fetched
+   * @param roleDefCache Cache to store role definition
```
jrhender:
```suggestion
   * @param roleDefCache Cache to store role definition. Cache is updated with role definition if not present.
```

jrhender (dismissed):
Looks good, thanks. I added some minor documentation suggestions as well which you can added if you think they are valuable.

jrhender (approved):

PR #84, opened 2022-10-07
feat: move did document resolution to resolvers

nichonien:
This PR consists of changes :
1. Remove issuance verification check from revocation status check
2. Add DIDDocument resolution to credential resolver.
3. Adds caching for DID Document

### DIFF
--- a/packages/vc-verification/src/resolver/Ipfs-credential-resolver.ts
+++ b/packages/vc-verification/src/resolver/Ipfs-credential-resolver.ts
@@ -5,9 +5,15 @@ import { Resolver } from '@ew-did-registry/did-ethr-resolver';
 import {
   RegistrySettings,
   IServiceEndpoint,
+  IDIDDocument,
 } from '@ew-did-registry/did-resolver-interface';
 import * as jwt from 'jsonwebtoken';
-import { RoleEIP191JWT, RolePayload } from '../models';
+import {
+  RoleEIP191JWT,
+  RolePayload,
+  IDIDDocumentCache,
+  IRoleCredentialCache,
+} from '../models';
 import {
   isEIP191Jwt,
   filterOutMaliciousClaims,
@@ -17,8 +23,7 @@ import {
 } from '../utils';
 import { CredentialResolver } from './credential-resolver';
 import { VerifiableCredential } from '@ew-did-registry/credentials-interface';
-import type { RoleCredentialSubject } from '@energyweb/credential-governance';
-import { IRoleCredentialCache } from '../models/cache-interfaces';
+import { RoleCredentialSubject } from '@energyweb/credential-governance';
 
 export class IpfsCredentialResolver implements CredentialResolver {
   private _ipfsStore: IDidStore;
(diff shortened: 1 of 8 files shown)


packages/vc-verification/src/verifier/revocation-verification.ts
```diff
@@ -238,14 +248,9 @@ export class RevocationVerification {
             await this._statusListEntryVerificaiton.verifyCredentialStatus(
               issuerCredential.credentialStatus
             );
-          } else {
-            const rolePayload =
-              await this.claimIssuerVerification.verifyIssuance(
-                issuer,
-                issuers?.roleName,
-                roleCredentialCache
-              );
```
jrhender: I think this is the implementation of point "1. Remove issuance verification check from revocation status check", right?
Could you explain how this affects the way in which the client trusts `StatusListCredential` @nichonien ? It seems to me this would open the possibility of the status list being inauthentic. Is it that the client is trusting that the server has done this issuance check and is serving a valid `StatusListCredential`?

jrhender (commented):
I would recommend splitting this PR into two, one that changes the DID document retrieval/caching and one that removes the issuer verification from the statuslist. The reason for this is that the changes could be applied independently and we may want to accept one change but not the other.

packages/vc-verification/src/verifier/revocation-verification.ts
nichonien: @jrhender We perform this validation (issuance) in the verifyIssuer call also (which is just followed by this revocation check - https://github.com/energywebfoundation/ew-credentials/blob/08461849f69eb288e276c45f874e052eff386047/packages/vc-verification/src/verifier/issuer-verification.ts#L104). So, having the same here in revocation is just redundant check and it is best to only check the status (expiration and revocation) in this part of the verification.

nichonien: Removed this change from this PR

JGiter (commented):
It is becoming more and more overcomplicated. I still beleive we need to separate caching from fetching

packages/vc-verification/src/resolver/Ipfs-credential-resolver.ts
```diff
@@ -225,4 +248,23 @@ export class IpfsCredentialResolver implements CredentialResolver {
...
     ).filter(isVerifiableCredential);
   }
+
+  /**
+   * Fetches DID Document for the given DID
+   * @param did subject DID
+   * @param didDocumentCache Cache to store DIDDocument. Cache is updated with Document retrieved for the DID
+   * @returns
+   */
+  async getDIDDocument(
```
jrhender, line 258: I think `getDIDDocument` could be private and then it doesn't need to be in the interface. Does this sound right?
```suggestion
  private async getDIDDocument(
```

nichonien: @jrhender We need this public, since it is used by `ClaimIssuerVerification` - https://github.com/energywebfoundation/ew-credentials/blob/853c4135777f19df46edea6de2b0d43ebe76d05e/packages/vc-verification/src/verifier/claim-issuer-verification.ts#L120

nichonien: And this would also allow PDA to pass cacheClient call inside getDIDDocument to fetch DIDDocument from SSI-Hub

jrhender: https://github.com/energywebfoundation/ew-credentials/blob/853c4135777f19df46edea6de2b0d43ebe76d05e/packages/vc-verification/src/verifier/claim-issuer-verification.ts#L120

At this line, why not use `this._issuerDefResolver` to get the `issuerDIDDoc`?
It is counterintuitive to me that the class that the `ClaimIssuerVerification` class has an `isserDefResolver` but doesn't use it to get the `issuerDIDDoc`.

nichonien: @jrhender I didn't get the correlation between the RoleDefinition resolver and DID Document. To me DID Document is more related to holder's credential. `this._issuerDefResolver` only resolves `RoleDefintion` from ENS.

jrhender: Oh you're right! A moment of confusion on my part 😅

jrhender (approved):
Looks good, thanks

PR #85, opened 2022-10-10
fix: remove issuance check from revocation status check

nichonien:


nichonien:
@jrhender Can we continue [this](https://github.com/energywebfoundation/ew-credentials/pull/84#discussion_r989984201) discussion here? 

I think it is completely fine to remove this check since it is being performed in `verifyIssuer` call as well.

jrhender:
> I think it is completely fine to remove this check since it is being performed in verifyIssuer call as well.

Thanks @nichonien . I'm not sure all possible paths are covered if we make this change. Here's a flow that I'm trying to work through:

1. start by calling `IssuerVerification.verifyIssuer` for `roleA`. It turns out that `roleA` issuance is authorized by `roleB`. `verifyIssuer` therefore retrieves the issuer's `roleB` credential and checks if it is revoked by calling `RevocationVerification.checkRevocationStatus`.
(`verifyIssuer` will verify issuance of the `roleB` credential, I'm with you on that 👍 )
2. `checkRevocationStatus` proceeds with checking the revocation status of the `roleB` credential. It turns out that `roleB` is marked as revoked and is revocable by `roleC`. So `checkRevocationStatus` retrieves the `roleC` credential of the DID which revoked the `roleB` credential. -> Now, at some point, we need to check if the issuer of the `roleC` credential was correct. I think this is the check that we might lose if we remove this code. Does this seem like the right assessment to you or not?

Maybe we need a test case for the above?

nichonien:
> 2. `checkRevocationStatus` proceeds with checking the revocation status of the `roleB` credential. It turns out that `roleB` is marked as revoked and is revocable by `roleC`. So `checkRevocationStatus` retrieves the `roleC` credential of the DID which revoked the `roleB` credential. -> Now, at some point, we need to check if the issuer of the `roleC` credential was correct. I think this is the check that we might lose if we remove this code. Does this seem like the right assessment to you or not?
> 
> Maybe we need a test case for the above?

@jrhender The above case is covered by `verifyRevoker` - https://github.com/energywebfoundation/ew-credentials/blob/98b4d13b3c29a1e3767d1e059431e8892efdbe64/packages/vc-verification/src/verifier/revocation-verification.ts#L169 method. So if `roleB` was revoked then [verifyRevoker](https://github.com/energywebfoundation/ew-credentials/blob/98b4d13b3c29a1e3767d1e059431e8892efdbe64/packages/vc-verification/src/verifier/revocation-verification.ts#L272) will be called and it will validate issuance for `roleC`.

jrhender:
@nichonien Ah, that's for pointing that out.
Yes, you're right. I see that verification of the `revoker` is done in the case that I described at this line: 

https://github.com/energywebfoundation/ew-credentials/blob/98b4d13b3c29a1e3767d1e059431e8892efdbe64/packages/vc-verification/src/verifier/revocation-verification.ts#L272

It seems we're missing a test for this case though? I think it would be very similar to this test case but where the revocation isn't actually done by the authorized party (i.e. the `manager` in this test case):
https://github.com/energywebfoundation/ew-credentials/blob/98b4d13b3c29a1e3767d1e059431e8892efdbe64/packages/vc-verification/test/vc-revocation-test.ts#L438

JGiter (approved):

nichonien:
> It seems we're missing a test for this case though? I think it would be very similar to this test case but where the revocation isn't actually done by the authorized party (i.e. the `manager` in this test case):

@jrhender we have this covered in the test as well : https://github.com/energywebfoundation/ew-credentials/blob/98b4d13b3c29a1e3767d1e059431e8892efdbe64/packages/vc-verification/test/issuer-verification-test.ts#L782

jrhender (approved):
Ah okay, thanks for bearing with me! Looks good to me then 👍

PR #86, opened 2022-11-01
feat: add transfer ownership script for claimmanager

nichonien:


packages/onchain-claims/contracts/Migrations.sol
```diff
@@ -0,0 +1,19 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.4.22 <0.9.0;
+
+contract Migrations {
```
JGiter, line 4: Just curious why we need this? I though that migrations are used to track deployment, but here we aren't deploying anything

nichonien: Yes, we are not deploying anything as of now. But for testing different use cases ( like transferring of ownership, we need to deploy again for testing ). So I though it would be good to keep it for future testing as well.

JGiter (dismissed):

JGiter (commented):
Please correct me if any of the following is wrong:
- Proxy is used to be able to upgrade ClaimManager
- After transferring proxy to safe we will be able to request updating of CM

JGiter (approved):

nichonien:
> Proxy is used to be able to upgrade ClaimManager

Yes, we interact with proxy contract for any execution and it delegates call to the implementation contract. So for upgrading or transferring we execute call from proxy.

>  After transferring proxy to safe we will not be able to request updating of CM

Well, we can upgrade / transfer through safe itself. For any major upgrade like adding of functionality, we need to do it through script and for that we would transfer the ownership for the moment to the developer working on it.

Does this answer your question?

JGiter:
Ok, thanks

PR #87, opened 2022-11-02
build: update @ew-did-registry@0.8.1-alpha.996.0

nichonien:


package.json
```diff
@@ -68,6 +68,7 @@
     "eslint-plugin-import": "2.25.4",
     "ganache-cli": "^6.12.2",
     "husky": "^4.3.8",
+    "keyv":"4.5.0",
```
JGiter, line 71: Was this added accidentially?

nichonien: This was added to resolve this error - https://github.com/energywebfoundation/ew-credentials/actions/runs/3369032554/jobs/5588238008#step:6:800

JGiter: Hmm, ok. There is also https://www.npmjs.com/package/@types/keyv

nichonien: I saw this as well, but it is deprecated so I thought not to use it.

JGiter (approved):

PR #88, opened 2022-11-03
fix: update imports to remove circular dependency

nichonien:


nichonien:
Warning for circular dependency was coming while build : https://github.com/energywebfoundation/ew-credentials/actions/runs/3374937651/jobs/5601088299#step:6:787

JGiter (approved):

PR #91, opened 2022-12-01
fix: downgrade did-ipfs-store to v0.7.1-alpha.816

nichonien:


jrhender:
A potential challenge with this approach is that then ssi-hub couldn't update to newer ew-credential versions, right? As it relies on the ew-did-reg version that supports IPFS cluster.

nichonien:
> A potential challenge with this approach is that then ssi-hub couldn't update to newer ew-credential versions, right? As it relies on the ew-did-reg version that supports IPFS cluster.

As of now, `sis-hub` doesn't support the cluster version which is v8 and the current version is v7.

jrhender (approved):
Seems good to me

PR #93, opened 2022-12-13
feat: add schema to field definition for json type

dawidgil:


nichonien (dismissed):

packages/credential-governance/src/types/domain-definitions.ts
```diff
@@ -44,6 +44,7 @@ export interface IFieldDefinition {
   maxValue?: number;
   minDate?: Date;
   maxDate?: Date;
+  schema?: string;
```
JGiter: Aren't we going to create schema with some of the libs, like https://ajv.js.org/? If so, then `schema` should have type used by this lib

dawidgil: Dunno, if you have capacity for it, you can make it working. 
@jrhender What do you think? Should we use it at this stage?

dawidgil: Anyway the more i think about this json schema the more i think we need a long term solution, not "patching" like this.

jrhender: > Anyway the more i think about this json schema the more i think we need a long term solution, not "patching" like this.

Does this comment of yours still stand? I think this addition of the `schema` property is roughly the solution that we are still going with, right?

jrhender: - I think a little documentation here couldn't hurt
- I think that using an object (represented by `Record<string, unknown>`) type, rather than a `string`, is better as it could allow us to conceivably change the type to the `ajv` type as @JGiter suggested without needing to migrate the role definitions.
```suggestion

  /**
   * A JSON Schema definition
   * For more information about JSON Schema, see https://json-schema.org/
   */
  schema?: Record<string, unknown>;
```

jrhender (commented):
I think it is better to use `Record<string, unknown>` rather than `string`

packages/credential-governance/src/types/domain-definitions.ts
dawidgil: so we do not use JSON.stringify for this case anymore? We will store real object here?

jrhender (approved):
Looks good to me

PR #102, opened 2023-02-23
fix: move dependencies to devdependencies

nichonien:


packages/vc-verification/package.json
```diff
@@ -59,8 +57,10 @@
     "ganache-cli": "^6.12.2",
     "ipfs": "^0.58.3",
     "ipfsd-ctl": "3.0.0",
+    "ipfs-http-client": "^43.0.0",
     "js-sha3": "^0.8.0",
     "jsonwebtoken": "^9.0.0",
+    "lodash": "^4.17.21",
```
jrhender: `lodash` is used here: https://github.com/energywebfoundation/ew-credentials/blob/a1e6ed6129485c2a4f33db4963ddb30684d144c2/packages/vc-verification/src/utils/formatters.ts
Is this doesn't seem to be only for tests?

jrhender (changes requested):

packages/vc-verification/package.json
nichonien: Aah, yes! I didn't see that. I will move it back to `dependencies`

packages/onchain-claims/package.json
```diff
@@ -58,6 +57,7 @@
     "@openzeppelin/contracts": "4.8.0",
     "@openzeppelin/contracts-upgradeable": "4.8.0",
     "@openzeppelin/truffle-upgrades": "1.17.0",
+    "@poanet/solidity-flattener": "^3.0.9",
```
artursudnik, line 60: this is misleading because the version is bumped here, but this dependency is not upgraded (no change to the lockfile). So, it's better to use `^3.0.7` here or keep `^3.0.9` and upgrade this package

nichonien: AFAIK only `package-lock` at root level needs to be maintained. There are mixed opinions on this topic, I will update as you said.

artursudnik: but, there is also a `package-lock.json` in this repository, so if we move to node v18 here, version 3.0.7 enforced by the lockfile will complain to install when developing locally or in CI

nichonien: These `package-lock` files are ignored and only root level is considered is what I have read.

artursudnik (approved):

PR #105, opened 2023-05-11
Optimize domain read

JGiter:
Current implementation of domain reading includes unnecessary alocations for intermediate results:
- for raw logs https://github.com/energywebfoundation/ew-credentials/blob/2867b93bda7264c7b36ea89b868a2b151ff34dd2/packages/credential-governance/src/domain-hierarchy.ts#L216
- decoded logs https://github.com/energywebfoundation/ew-credentials/blob/2867b93bda7264c7b36ea89b868a2b151ff34dd2/packages/credential-governance/src/domain-hierarchy.ts#L218

This PR streamifies logs processing meaning that memory allocated only for logs that are used for currently parsed domain. With this change memory usage decreases from 750 mb to 230 mb

The level of simultaneously processed logs is determined by `concurrency` parameter

I am uncertain about optimal concurrency. The bigger the value, the less requests to rpc will be made and synchronization will be faster. But from other side more memory will be allocated for logs and processing tasks. 

Some test results:

| log batch size | number of parallel reads | event  |  time | max mem, mb |
|--------------------|----------------------------------|-------------|--------|---------------|
| Infinity | Infinity | TextChanged | 13.740s |  487 | 
| Infinity | Ininity |  DomainUpdated | 3:52.032 | 881 |
| Infinity | 10 |  TextChanged | 2:45.623 | 240 |
| Infinity | 10 |  DomainUpdated | 49:20.397 | 736 |
| Infinity | 100 |  TextChanged | 23.487s | 343 |
| Infinity | 100 |  DomainUpdated | 7:16.738 | 1042 |
| 25_000 | 100 |  TextChanged | 8:36.913 | 242 |
| 25_000 | 100 |  DomainUpdated | 12:33.312 | 574 |
| 10_000 | 100 | TextChanged | 21:08.535 | 240 |
| 10_000 | 100 | DomainUpdated | 25:13.014 | 589 |
...

### DIFF
--- a/packages/credential-governance/src/domain-hierarchy.ts
+++ b/packages/credential-governance/src/domain-hierarchy.ts
@@ -206,7 +206,6 @@ export class DomainHierarchy {
   };
 
   private getDomainsFromLogs = async ({
-    provider,
     parser,
     event,
     contractInterface,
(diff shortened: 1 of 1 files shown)


packages/credential-governance/src/domain-hierarchy.ts
```diff
@@ -214,35 +216,53 @@ export class DomainHierarchy {
...
+        if (mem > maxMem) {
+          maxMem = mem;
+          process.stdout.write(`max mem ${maxMem / 1e6} mb`);
+          process.stdout.clearLine(0);
+          process.stdout.cursorTo(0);
+        }
+
+        return parsed;
+      },
+      { concurrency: 3 }
```
nichonien: @JGiter Is there any reason to set `concurrency` to `3` here, the default concurrency is `infinity` for `pMap`,  I have tested it with different `concurrency` value and the best outcome came with the default value. What do you think?

JGiter: I didn't decided yet how to estimate `concurrency`, so I set it arbitrary for now. I changed it from default, because if I understand correctly we are using `pMap` in order to control number of running tasks. 
May be it is fine to not limit concurrency, if the real memory spike cause is indeed intermediate variables and not number of Promise's. 
I will try different cases to determine optimal concurrency or whether we need to limit concurrency at all

nichonien: @JGiter I tested the latest changes and it consumes ~250 mbs which seems to be an efficient mem consumption. I think we should merge and release this change, test locally with SSI-Hub and if we don't see any spike then we can push it to dev and later to staging.

I will remove the console logs to get the tests running.

JGiter: Yes lets try this

packages/credential-governance/src/domain-hierarchy.ts
```diff
@@ -214,35 +215,44 @@ export class DomainHierarchy {
...
+    let maxMem = mem;
+    console.log(`max mem ${maxMem / 1e6} mb`);
+
+    mem = process.memoryUsage().heapUsed;
+    if (mem > maxMem) {
+      maxMem = mem;
+      console.log(`mem ${maxMem / 1e6} mb`);
+    }
+
+    const logBatchSize = 25_000;
```
nichonien:
```suggestion
    const logBatchSize = 25000;
```
@JGiter Is this intentional or by mistake? To run the script locally I changed it to the suggested value else it wasn't ending for me.

nichonien (approved):

packages/credential-governance/src/domain-hierarchy.ts
JGiter: It was intentional. Such syntax makes big number more readable in my opinion. But I am ok with your suggestion

PR #107, opened 2023-05-15
add script to read domain

JGiter:


nichonien:
@JGiter IMO we should have these scripts for future debugging and testing.

JGiter:
ok, I changed script a bit https://github.com/energywebfoundation/ew-credentials/pull/108

PR #110, opened 2023-06-20
Downgrade did-ipfs-store@0.7.1-alpha.816.0

JGiter:


packages/vc-verification/test/claim-issuer-verification.ts
```diff
@@ -128,17 +126,11 @@ export function claimIssuerVerificationTests(): void {
     managerDid = `did:${Methods.Erc1056}:${Chain.VOLTA}:${managerAddress}`;
     manager = EwSigner.fromPrivateKey(managerKeys.privateKey, providerSettings);
 
-    verifierKeys = new Keys({
```
JGiter, line 131: Not used

nichonien (approved):

PR #132, opened 2025-01-13
feat/resolve ipfs by multi cid

JGiter:
Several changes to help resolve claim from Ipfs:
* will try to resolve by CID_V0 and CID_V1
* set resolving timout

If claim still not resolved, it will be skipped


JGiter:
Replaced by https://github.com/energywebfoundation/ew-credentials/pull/133

PR #133, opened 2025-01-14
feat/skip not resolved token

JGiter:
Try to resolve token in IPFS within a given timeout. If token is not resolved it is skipped similarly as invalid service endpoints


jrhender (approved):

PR #138, opened 2025-04-09
bug/no-ipfs-timeout

JGiter:
Two bug fixes:
* skipping resolution from IPFS on timeout
* requests to IPFS are sent one by one instead of all together. This is needed, because IPFS refuses to serve too many requests

### DIFF
--- a/packages/vc-verification/src/resolver/Ipfs-credential-resolver.ts
+++ b/packages/vc-verification/src/resolver/Ipfs-credential-resolver.ts
@@ -193,7 +193,7 @@ export class IpfsCredentialResolver implements CredentialResolver {
   ): Promise<RoleEIP191JWT[]> {
     const didDocument = await this.getDIDDocument(did, didDocumentCache);
     const services: IServiceEndpoint[] = didDocument.service || [];
-    let resolved: Array<RoleEIP191JWT> = [];
+    const resolved: Array<RoleEIP191JWT> = [];
     for (const { serviceEndpoint } of services) {
       if (!isCID(serviceEndpoint)) {
         continue;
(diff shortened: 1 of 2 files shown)


packages/vc-verification/src/resolver/Ipfs-credential-resolver.ts
```diff
@@ -193,34 +193,32 @@ export class IpfsCredentialResolver implements CredentialResolver {
...
     const didDocument = await this.getDIDDocument(did, didDocumentCache);
     const services: IServiceEndpoint[] = didDocument.service || [];
-    return (
-      await Promise.all(
-        services.map(async ({ serviceEndpoint }) => {
-          if (!isCID(serviceEndpoint)) {
-            return {};
-          }
+    const resolved: Array<RoleEIP191JWT> = [];
+    for (const { serviceEndpoint } of services) {
```
kim-energyweb, line 197: Performance issue. Consider Promise.allSettled().

packages/vc-verification/src/resolver/Ipfs-credential-resolver.ts
```diff
@@ -238,30 +236,29 @@ export class IpfsCredentialResolver implements CredentialResolver {
...
-          // expect that JWT would have 3 dot-separated parts, VC is non-JWT credential
-          if (!(credential.split('.').length === 3)) {
-            vc = JSON.parse(credential);
-          }
-          return vc as VerifiableCredential<RoleCredentialSubject>;
-        })
-      )
-    ).filter(isVerifiableCredential);
+    const resolved: Array<VerifiableCredential<RoleCredentialSubject>> = [];
+    for (const { serviceEndpoint } of services) {
```
kim-energyweb, line 240: Performance issue

kim-energyweb (changes requested):

packages/vc-verification/src/resolver/Ipfs-credential-resolver.ts
JGiter: this is for purpose. When there are too many claims ( I tested with 30), then all these requests are sent at once and IPFS returs error something like `too many p2p requests`

kim-energyweb: How about introducing `p-limit` or similar approach?

kim-energyweb (changes requested):

kim-energyweb (approved):

PR #139, opened 2025-04-14
IPFS resolve batch

JGiter:


packages/vc-verification/package.json
```diff
@@ -45,7 +45,8 @@
     "@ew-did-registry/did-store-interface": "^0.8.1-alpha.1122.0",
     "@ew-did-registry/revocation": "^0.8.1-alpha.1122.0",
     "ethers": "^5.7.2",
-    "lodash": "^4.17.21"
+    "lodash": "^4.17.21",
+    "p-limit": "3"
```
JGiter, line 49: latest is v6, but apparently it is not compatible with some other dependencies or Node, because I got this in CI https://github.com/energywebfoundation/ew-credentials/actions/runs/14440724722/job/40490070736

kim-energyweb (approved):

